PR #15, opened 2018-06-28
Agr 1031

cmpich:
Create PhenotypeAnnotationDocument docs in ES from the phenotype in neo4j...

### DIFF
--- a/agr_java_core/src/main/java/org/alliancegenome/core/translators/document/PhenotypeTranslator.java
+++ b/agr_java_core/src/main/java/org/alliancegenome/core/translators/document/PhenotypeTranslator.java
@@ -1,13 +1,13 @@
 package org.alliancegenome.core.translators.document;
 
 import org.alliancegenome.core.translators.EntityDocumentTranslator;
-import org.alliancegenome.es.index.site.doclet.PublicationDoclet;
-import org.alliancegenome.es.index.site.doclet.SourceDoclet;
-import org.alliancegenome.es.index.site.doclet.SpeciesDoclet;
+import org.alliancegenome.core.translators.doclet.PublicationDocletListTranslator;
 import org.alliancegenome.es.index.site.document.PhenotypeAnnotationDocument;
 import org.alliancegenome.es.index.site.document.PhenotypeDocument;
-import org.alliancegenome.neo4j.entity.SpeciesType;
-import org.alliancegenome.neo4j.entity.node.*;
+import org.alliancegenome.neo4j.entity.node.Feature;
+import org.alliancegenome.neo4j.entity.node.Gene;
+import org.alliancegenome.neo4j.entity.node.Phenotype;
+import org.alliancegenome.neo4j.entity.node.PhenotypeEntityJoin;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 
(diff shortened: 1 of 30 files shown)


agr_java_core/src/main/java/org/alliancegenome/core/translators/document/PhenotypeTranslator.java
```diff
@@ -0,0 +1,281 @@
...
+                                groupingBy(join -> Optional.ofNullable(join.getFeature())))
+                );
+
+        // sort by gene symbol
+        return geneMap.entrySet().stream()
+                .sorted(Map.Entry.comparingByKey())
+                .collect(toMap(Map.Entry::getKey, Map.Entry::getValue));
+    }
+
+    private PublicationDoclet getPublicationDoclet(PhenotypeEntityJoin association, Publication publication) {
```
oblodgett: Can this be turned into a Doclet Translator this code also exists in the DiseaseTranslator.java?

agr_java_core/src/main/java/org/alliancegenome/core/translators/document/PhenotypeTranslator.java
```diff
@@ -0,0 +1,281 @@
...
+            }
+        }
+        return pubDocletList;
+    }
+
+    private SpeciesDoclet getSpeciesDoclet(Gene gene) {
+        return gene.getSpecies().getType().getDoclet();
+    }
+
+    public Iterable<PhenotypeAnnotationDocument> translateAnnotationEntities(List<Phenotype> phenotypeList) {
```
oblodgett: Can translateAnnotationEntities for PhenotypeAnnotationDocument be turned into its own translator so this class won't be so large?

agr_java_core/src/main/java/org/alliancegenome/core/translators/document/PhenotypeTranslator.java
```diff
@@ -0,0 +1,281 @@
...
+        document.setPrimaryId(phenotype.getPrimaryKey());
+        document.setPhenotypeStatement(phenotype.getPhenotypeStatement());
+        if (shallow)
+            return document;
+
+        return document;
+    }
+
+    private List<SourceDoclet> getSourceUrls(DOTerm doTerm) {
+
```
oblodgett: This code is duplicate from the DiseaseTranslator.java shouldn't this be put into one place?

agr_java_core/src/main/java/org/alliancegenome/core/translators/document/PhenotypeTranslator.java
cmpich: In the DiseaseTranslator it uses DiseaseEntityJoin instead of PhenotypeEntityJoin... It also uses the DOTerm entity instead of Phenotype.
I could create a super class called EntityJoin that contains gene and feature and publication and would leave the child class just adding the main entity - DoTerm or Phenotype...

oblodgett: In this method, only the publication is used. I was suggesting that Publication needs to be a translator. Plus why is association passed in and not used? Maybe the evidences codes could be passed in as a list?

agr_java_core/src/main/java/org/alliancegenome/core/translators/document/PhenotypeTranslator.java
cmpich: This method is not used in the PhenotypeTranslator. Removing it altogether.

PR #47, opened 2018-08-02
Agr 857

cmpich:
Create orthology web service endpoint for use case
retrieve orthology data for a given gene.


kevinschaper:
Thanks for adding more tests!

PR #51, opened 2018-08-15
fix table formatting in "Data Type" section

gildossantos:


gildossantos:
I'm catching a few other minor typos so I'll try and fix all that before next PR.

PR #62, opened 2018-08-23
fix compiler errors

cmpich:


oblodgett:
Why does this cause compiler errors?

PR #88, opened 2018-09-12
add annotationCount variable on group level

cmpich:


agr_api/src/test/java/org/alliancegenome/api/GeneTest.java
```diff
@@ -123,7 +123,7 @@ public void checkOrthologyForTwoSpecies() throws IOException {
     public void checkOrthologyForSingleSpecies() throws IOException {
 
         OrthologyController controller = new OrthologyController();
-        String responseString = controller.getSingleSpeciesOrthology("10090", "stringent", null, 20, 1);
+        String responseString = controller.getSingleSpeciesOrthology("559292", "stringent", "OMA", 20, 1);
```
oblodgett: I know this is just a test file, but shouldn't we use:

SpeciesType.MOUSE.getTaxonIDPart() in order to make sure we are not using strings everywhere.

PR #129, opened 2018-10-09
AGR-1196  Fix cellular component expression indexing for genes

kevinschaper:
@cmpich  We had a collision on both adding a p12, so I thought it might be time to start moving towards meaningful names to the match clauses.


agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/Ontology.java
```diff
@@ -1,6 +1,10 @@
 package org.alliancegenome.neo4j.entity.node;
 
 import org.alliancegenome.neo4j.entity.Neo4jEntity;
+import org.neo4j.ogm.annotation.Relationship;
```
cmpich: Are these imports needed?

PR #163, opened 2018-11-02
AGR-1271 add crossReference info for features from new node.

cmpich:


oblodgett:
are you sure you want this?
`query += " OPTIONAL MATCH featureCrossRef1=(ff:Feature)--(crossRef1:CrossReference)";`
that will match ALL Features

I think you might want (ff) to be (s)
or (g:Gene)--(ff:Feature)--(etc…)

PR #183, opened 2019-02-04
Update Submission_README.md

cmpich:
Added phenotype, expression and interaction data type info.


oblodgett:
The booleans for INTERACTION are wrong... should be false | false | false. Anyway in talking to @sierra-moxon on friday, I am going to make a change to the system to support Type and Subtype... this makes things more generic, and more aligns with the loader and gives us the flexibility to do stuff like
Type: Ontology Subtype: GO 
Type: BGI, Subtype: MGI
Type: Interaction: Subtype: Interaction

PR #185, opened 2019-02-06
Fixed disease defLinks mapping

kevinschaper:


oblodgett:
can this be fixed on the loader side...

kevinschaper:
Whoops!  Good that it's getting fixed at the source

PR #186, opened 2019-02-06
fixed unit tests

cmpich:


cmpich:
I will modify some of the tests when we are farther along to make them less volatile.

PR #188, opened 2019-02-11
AGR-1403 AGR-1402 add name to JSON output for phentoypes

cmpich:


cmpich:
Modified as discussed with just the Views...

PR #202, opened 2019-02-28
AGR-1343 - Allele indexing fixes 

kevinschaper:


oblodgett:
See indexer channel in slack.

PR #213, opened 2019-03-13
Attempt to catch all exceptions

oblodgett:


oblodgett:
Based on this thread:
http://www.drdobbs.com/jvm/uncaught-java-thread-exceptions/240148320

PR #231, opened 2019-04-23
Upgraded Wildfly to Thorntail

oblodgett:


oblodgett:
Because the upgrade,  upgraded maven which is not compatible to the old version: https://stackoverflow.com/questions/55117508/problem-executing-goal-wildfly-swarm-plugin-api-incompatibility-java-lang-abst/55118738#55118738

PR #249, opened 2019-05-17
AGR-1579 adjust sections/categories section in the JSON output stream…

cmpich:
… as requested


cmpich:
Moved all relevant classes into the entity package for consistency reasons.

PR #256, opened 2019-05-24
AGR-1579 add 'All xxx' section for each category as per request from …

cmpich:
…the WG


agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/GeneRepository.java
```diff
@@ -95,105 +95,6 @@ public Gene getOneGeneBySecondaryId(String secondaryIdPrimaryKey) {
         return null;
     }
 
-    public List<Allele> getAlleles(String primaryKey) {
```
cmpich: not used any longer....

PR #262, opened 2019-05-29
Fixes for allele sorting and disease ribbon 

cmpich:


cmpich:
Also, add genes in the disease ribbon that do not have any disease found.

PR #282, opened 2019-06-17
AGR-1637

kevinschaper:
I want to see the automated test output on this, the fix is working locally but even before wasn't showing the same problem that stage has.


kevinschaper:
oops, wrong branch..

PR #283, opened 2019-06-17
AGR-1637  Tweaked allele related data link to use name rather than nameKey

kevinschaper:
Not sure I want this merged yet, mostly I want to see the automated test output to compare with odd difference between local site and staging.


kevinschaper:
This test is passing for me locally, but this failure is the confirmation I was looking for that there was a difference.

```
sa2545 has gene alleles related data links(org.alliancegenome.api.RelatedDataServiceIntegrationSpec)  Time elapsed: 0.096 sec  <<< FAILURE!
org.spockframework.runtime.ConditionNotSatisfiedError: 
Condition not satisfied:

relatedDataLink.count <= max
|               |     |  |
|               5     |  1
|                     false
org.alliancegenome.es.model.search.RelatedDataLink@5793b87

```

PR #286, opened 2019-06-18
AGR-1611 enable orthologyGenes to be included in JSON output for dise…

cmpich:
…ase ribbon table


cmpich:
I also made changes to throw a RestException when cache is still being processed and data are not available.

PR #298, opened 2019-08-01
AGR-1679 refactor expression ribbon summary to behave in the same way…

cmpich:
… as the disease ribbon


cmpich:
Renamed the classes for the ribbon to not disease specific any longer so I can re-use them for the expression ribbon.

cmpich:
I checked each file and it looks fine to me now. Yes, Intellij refactor has issues.

PR #314, opened 2019-09-13
AGR-1201

prmani:
Changes to Phenotype endpoint to accept a list of genes


agr_api/src/main/java/org/alliancegenome/api/service/GeneService.java
```diff
@@ -58,20 +59,21 @@ public Gene getById(String id) {
...
         return response;
     }
 
-    public EntitySummary getPhenotypeSummary(String geneID) {
+    public EntitySummary getPhenotypeSummary(List<String> geneIDs) {
+
         EntitySummary summary = new EntitySummary();
-        summary.setNumberOfAnnotations(phenoRepo.getTotalPhenotypeCount(geneID, new Pagination()));
-        summary.setNumberOfEntities(phenoRepo.getDistinctPhenotypeCount(geneID));
+        geneIDs.forEach(geneID ->summary.setNumberOfAnnotations(phenoRepo.getTotalPhenotypeCount(geneID, new Pagination())));
```
cmpich: It looks like this lambda loop is just overwriting each loop element's previous call, i.e. the setting of the number is set to various numbers where only the last call will be persisted.

PR #322, opened 2019-09-25
agm   (AGR-1687)

cmpich:
Fixed unit tests.

### DIFF
--- a/agr_api/src/main/java/org/alliancegenome/api/rest/interfaces/DiseaseRESTInterface.java
+++ b/agr_api/src/main/java/org/alliancegenome/api/rest/interfaces/DiseaseRESTInterface.java
@@ -7,7 +7,6 @@ import io.swagger.annotations.ApiOperation;
 import io.swagger.annotations.ApiParam;
 import org.alliancegenome.core.service.JsonResultResponse;
 import org.alliancegenome.neo4j.entity.DiseaseAnnotation;
-import org.alliancegenome.neo4j.entity.PrimaryAnnotatedEntity;
 import org.alliancegenome.neo4j.entity.node.DOTerm;
 import org.alliancegenome.neo4j.view.View;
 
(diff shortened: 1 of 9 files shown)


agr_api/src/main/java/org/alliancegenome/api/rest/interfaces/DiseaseRESTInterface.java
```diff
@@ -134,9 +187,35 @@
...
+            @QueryParam("filter.disease") String disease,
+            @ApiParam(value = "filter by source")
+            @QueryParam("filter.source") String source,
+            @ApiParam(value = "filter by reference")
+            @QueryParam("filter.reference") String reference,
+            @ApiParam(value = "filter by evidence code")
+            @QueryParam("filter.evidenceCode") String evidenceCode,
+            @ApiParam(value = "ascending order: true or false", allowableValues = "true,false", defaultValue = "true")
+            @QueryParam("asc") String asc);
+
```
oblodgett: Just wondering if maybe a better way to handle this instead of having many parameters we could do something like
List<String> filter_fields and List<String> filter_values and the URL string would look like:
`/endpoint?filter_fields=[disease,geneName]&filter_values=[diabetes,pax6]`

oblodgett: Then we could just have a util that takes both those lists and returns a map that is ready to go for filtering. That was we could just put those two params on all endpoints to get filtering. I know we would loose the documentability of each field, but would gain not having to change things if we wanted to have a new way to filter.

cmpich: I know and I have been hoping for something like this but then Swagger will not show each individual parameter.

cmpich: It's just more in your face, i.e. the parameters are just more explicit.

PR #329, opened 2019-10-01
AGR-1364 hard-code AO term order in cvs file for expression Location …

cmpich:
…ribbon

### DIFF
--- a/agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/GeneRepository.java
+++ b/agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/GeneRepository.java
@@ -10,10 +10,7 @@ import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.neo4j.ogm.model.Result;
 
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.InputStreamReader;
-import java.io.Serializable;
+import java.io.*;
 import java.net.URL;
 import java.util.*;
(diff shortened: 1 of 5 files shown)


agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/GeneRepository.java
```diff
@@ -533,23 +530,48 @@ private String getAllGenesQuery(OrthologyFilter filter) {
...
+        try {
+            String str = null;
+            in = this.getClass().getClassLoader().getResourceAsStream("anatomy-term-order.csv");
+            if (in != null) {
+                reader = new BufferedReader(new InputStreamReader(in));
+                while ((str = reader.readLine()) != null) {
+                    String[] token = str.split("\t");
+                    aoTermList.put(token[0], token[1]);
+                }
+            }
```
oblodgett: @cmpich can we put this code into the ConfigHelper file... we might need to do more of this kind of thing in the future. Something like `aoTermList = ConfigHelper.getAnatomyTermOrderList()` that way the paths can be handled there also.

oblodgett: Also can we keep ALL class variables at the top of the class... so that way they are all in one place and don't have to go hunting to find them?

cmpich: Yes. I will need it for the GO list as well.

PR #332, opened 2019-10-04
fix Phenotypes

cmpich:


cmpich:
The cypher query had a where clause that needed to be dropped.

PR #333, opened 2019-10-04
AGR-1709: fixed method getOrthologyMultiGeneJson to filter out orthol…

prmani:
…ogs based on stringencyFilter. default=stringent.

### DIFF
--- a/agr_java_core/src/main/java/org/alliancegenome/core/service/OrthologyService.java
+++ b/agr_java_core/src/main/java/org/alliancegenome/core/service/OrthologyService.java
@@ -170,15 +170,37 @@ public class OrthologyService {
     public static JsonResultResponse<OrthologView> getOrthologyMultiGeneJson(List<String> geneIDs, OrthologyFilter filter) {
         GeneCacheRepository repo = new GeneCacheRepository();
         List<OrthologView> orthologViewList = repo.getAllOrthologyGenes(geneIDs);
+        List<OrthologView> orthologViewFiltered = orthologViewList;
+        List<OrthologView> orthologViewFilteredModerate = orthologViewList;
 
         System.out.println("Number of genes for orthology: " + geneIDs.size());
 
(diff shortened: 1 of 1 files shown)


agr_java_core/src/main/java/org/alliancegenome/core/service/OrthologyService.java
```diff
@@ -202,15 +224,19 @@ private static void putGeneInfo(Map<String, Object> map, Gene gene) {
...
-    public static JsonResultResponse<OrthologView> getOrthologyGenes(List<String> geneIDList, OrthologyFilter filter) {
+    public static JsonResultResponse<OrthologView> getOrthologyGenes(List<String> geneIDList, OrthologyFilter orthoFilter) {
         GeneCacheRepository repo = new GeneCacheRepository();
         List<OrthologView> orthologViewList = repo.getAllOrthologyGenes(geneIDList);
         List<OrthologView> filteredOrthologViewList = orthologViewList;
-        if (filter.getStringency() != null && !filter.getStringency().equals(OrthologyFilter.Stringency.ALL)) {
-            filteredOrthologViewList = orthologViewList.stream()
-                    .filter(orthologView -> orthologView.getStringencyFilter().equalsIgnoreCase(filter.getStringency().name()))
+
+        System.out.println("Filter: " + orthoFilter.getStringency().name());
```
oblodgett: We use a logger for outputting info from the applications... please change the System.out.println -> log.info or log.debug thanks.

oblodgett: In order to get a logger into the class just add the following annotation to the class:

`@Log4j2`
`public class OrthologyService {`

prmani: I had used that merely for my own debugging purposes ..I took it out (but accidentally merged my pr :(

oblodgett: Its all good...

PR #335, opened 2019-10-04
AGR-1689  Index AffectedGenomicModels

kevinschaper:


agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/ModelIndexerRepository.java
```diff
@@ -0,0 +1,93 @@
...
+    public Map<String, Set<String>> getDiseaseMap(String species) {
+        String query = "MATCH (species:Species)-[:FROM_SPECIES]-(model:AffectedGenomicModel)--(disease:DOTerm)";
+        query += getSpeciesWhere(species);
+        query += " RETURN model.primaryKey as id, disease.name as value";
+
+        return getMapSetForQuery(query, getSpeciesParams(species));
+    }
+
+    public Map<String, Set<String>> getGeneMap(String species) {
+        String query = "MATCH (species:Species)-[:FROM_SPECIES]-(model:AffectedGenomicModel)-[:MODEL_COMPONENT|:SEQUENCE_TARGETING_REAGENT]-(feature)--(gene:Gene)";
```
cmpich: Sierra told me, that there is no one-to-one AGM-Gene relationship, only if we have only clean data. In the API I am not making a relationship between AGM on the DiseaseEntityJoin node and Gene nodes. Only on the disease page which shows all AGMs. The AGM section on the gene page I don't include them. Just curious if that is your understanding too?

PR #338, opened 2019-10-08
AGR-1711 Added variant consequence facet to alleles

kevinschaper:


agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleIndexerRepository.java
```diff
@@ -135,4 +138,12 @@ public AlleleDocumentCache getAlleleDocumentCache(String species) {
         return getMapSetForQuery(query, "a.primaryKey", "term.name", getSpeciesParams(species));
     }
 
+    public Map<String, Set<String>> getMolecularConsequence(String species) {
+        String query = "MATCH (species:Species)--(:Gene)-[:IS_ALLELE_OF]-(a:Allele)--(v:Variant)--(consequence:GeneLevelConsequence) ";
```
oblodgett: Depending on the speed of this query you might want to put all the relationships in it will help speed it up if there are issues.

PR #349, opened 2019-10-18
AGR-1753 added Phenotypestatement ordering to default order

prmani:
added another ordering to defaultOrder for alleles


agr_java_core/src/main/java/org/alliancegenome/core/service/ModelAnnotationsSorting.java
```diff
@@ -7,7 +7,7 @@
 import java.util.Comparator;
 import java.util.List;
 
-public class ModelAnnotationsSorting implements Sorting<DiseaseAnnotation> {
+public class  ModelAnnotationsSorting implements Sorting<DiseaseAnnotation> {
```
cmpich: If you removed this white space then this class would not have a single change. That would be better as it should not be changed for this bug anyway.

PR #358, opened 2019-10-23
AGR-1796, AGR-1797 suppress Allele type of models in the models table

cmpich:


cmpich:
Adding a helper cacher class to the mix which helps debugging disease-type caching issues. If it takes too long I can remove it again.

PR #362, opened 2019-10-28
AGR-1801 hid download link

prmani:


oblodgett:
@prmani would like to discuss this... we need to be able to have these visible in the build systems in order to test them, and I agree have them turned off in production... so anyway would like to discuss this before it goes into code.

cmpich:
What do you mean by 'visible in the build system'. Those endpoints should not show up on the swagger UI page. In prod and anywhere else.

oblodgett:
How does the UI test the endpoint? If its not available anywhere... I mean you would have to craft the URL by hand which is doable, but it would be much easier to know what params you have available to you. I am suggesting that somehow these endpoints be "turned on" on build so that they can be tested in the swagger UI. And be "turned off" in production so that they are not visible to the public.

cmpich:
There is no testing to be done on the disabled endpoints on stage. Once you merge the code to master (build) we should enable them again as they need to be fixed in the next release.

PR #395, opened 2019-12-19
AGR-1865 AGR-1877

cmpich:


agr_api/pom.xml
```diff
@@ -241,6 +241,14 @@
 					</includes>
 				</configuration>
 			</plugin>
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<configuration>
+					<source>11</source>
```
oblodgett: @cmpich can you double check that this is done on all the sub projects?

agr_java_core/src/main/java/org/alliancegenome/core/translators/tdf/DiseaseAnnotationToTdfTranslator.java
```diff
@@ -134,57 +131,107 @@ public String getAllRowsForModel(List<DiseaseAnnotation> diseaseAnnotations) {
...
+                                    row.setGeneticEntityID(entity.getId());
+                                    row.setGeneticEntityName(entity.getName());
+                                    row.setGeneticEntityType(entity.getType().getDisplayName());
+                                    row.setSpeciesID(annotation.getGene().getSpecies().getPrimaryKey());
+                                    row.setSpeciesName(annotation.getGene().getSpecies().getName());
+                                    row.setAssociation(annotation.getAssociationType());
+                                    row.setDiseaseID(annotation.getDisease().getPrimaryKey());
+                                    row.setDiseaseName(annotation.getDisease().getName());
+                                    row.setSource(annotation.getSource().getName());
+
```
prmani: the reference needs to be set here as well

PR #401, opened 2020-01-15
AGR-1899 Add support for finding features by hgvs name

kevinschaper:
I had some trouble because it turned out that <> aren't chars that Lucene needs to have escaped, but we were escaping them. I also tweaked the part of our ES query that adds a 10x boost for each term matched so that it doesn't further tokenize the matches. I'm not yet 100% confident about what side effects that change might have, but I'll have another PR coming in behind this one with some test infrastructure tweaks and that's where I'll be able to see any problems.

### DIFF
--- a/agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleIndexerRepository.java
+++ b/agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleIndexerRepository.java
@@ -63,6 +63,9 @@ public class AlleleIndexerRepository extends Neo4jRepository {
         log.info("Building allele -> phenotype statements map");
         alleleDocumentCache.setPhenotypeStatements(getPhenotypeStatementsMap(species));
 
+        log.info("Building allele -> variant name map");
+        alleleDocumentCache.setRelatedVariants(getRelatedVariants(species));
+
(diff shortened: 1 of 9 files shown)


agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleIndexerRepository.java
```diff
@@ -130,6 +133,14 @@ public AlleleDocumentCache getAlleleDocumentCache(String species) {
         return getMapSetForQuery(query, "a.primaryKey", "phenotype.phenotypeStatement", getSpeciesParams(species));
     }
 
+    public Map<String, Set<String>> getRelatedVariants(String species) {
+        String query = "MATCH (species:Species)--(:Gene)-[:IS_ALLELE_OF]-(a:Allele)--(v:Variant) ";
```
oblodgett: Might want to specify the relationship between `(a:Allele)--(v:Variant)` don't know if this query has performance issues or not?

kevinschaper: Doesn't look like it does, seems to complete in about a second happily.

PR #494, opened 2020-04-23
AGR-1812 AGR-1995 

cmpich:


cmpich:
Yes, the conflict you will get is minor. I think you created a new class for the devtool endpoints. I can help with the merge when it comes to it.

PR #495, opened 2020-04-23
Infinispan refactor

oblodgett:


agr_api/src/main/java/org/alliancegenome/api/rest/interfaces/CacheRESTInterface.java
```diff
@@ -25,9 +33,14 @@
 
 
     @GET
-    @Path("/{cacheSpace}/{id}")
+    @Path("/{cache_name}/{id}")
```
cmpich, line 36: Can we use camel case convention?

agr_api/src/main/java/org/alliancegenome/api/rest/interfaces/CacheRESTInterface.java
```diff
@@ -25,9 +33,14 @@
...
+    @ApiOperation(value = "Get Cache Object")
+    @JsonView(value = {View.Default.class})
     String getCacheObject(
+            @ApiParam(name = "id", value = "Search for an object by ID")
             @PathParam("id") String id,
-            @PathParam("cacheSpace") String cacheName
+            @ApiParam(name = "cache_name", value = "Cache Type to Search by")
+            @PathParam("cache_name") String cache_name
```
cmpich, line 43: camel case?

agr_java_core/src/main/java/org/alliancegenome/cache/CacheService.java
```diff
@@ -0,0 +1,138 @@
...
+        RemoteCache<String, String> remoteCache = manager.getCache(cache.getCacheName());
+        
+        if(remoteCache == null) {
+            return CacheConfig.createCache(cache);
+        }
+        
+        return remoteCache;
+    }
+
+    public <O> List<O> getCacheEntries(String entityID, CacheAlliance cacheSpace, Class<O> clazz) {
```
cmpich, line 66: What does the '0' in the generic syntax mean?

agr_java_core/src/main/java/org/alliancegenome/core/config/CacheConfig.java
```diff
@@ -0,0 +1,81 @@
...
+
+import lombok.extern.log4j.Log4j2;
+
+@Log4j2
+public class CacheConfig {
+
+    private static RemoteCacheManager manager;
+    
+    public static RemoteCacheManager defaultRemoteCacheManager() {
+        if(manager != null) return manager;
```
cmpich, line 18: Does that need to be made thread-safe? I don't know how log it takes to create the manager object...

agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/DiseaseAnnotation.java
```diff
@@ -1,18 +1,34 @@
...
+import java.util.stream.Collectors;
+
+import org.alliancegenome.api.entity.PresentationEntity;
+import org.alliancegenome.neo4j.entity.node.AffectedGenomicModel;
+import org.alliancegenome.neo4j.entity.node.Allele;
+import org.alliancegenome.neo4j.entity.node.CrossReference;
+import org.alliancegenome.neo4j.entity.node.DOTerm;
+import org.alliancegenome.neo4j.entity.node.ECOTerm;
+import org.alliancegenome.neo4j.entity.node.Gene;
+import org.alliancegenome.neo4j.entity.node.Publication;
```
cmpich, line 19: Why are you listing all the classes individually instead of using the 
import org.alliancegenome.neo4j.entity.node.*;
statement? is that an Eclipse thing?
I wished the import statement changes would not all change

agr_api/src/main/java/org/alliancegenome/api/rest/interfaces/CacheRESTInterface.java
oblodgett: Its only seen here in the API file but sure... i'll change it.

agr_java_core/src/main/java/org/alliancegenome/cache/CacheService.java
oblodgett: It's used to specify the object that will be used in the params.... this way you don't have to use `<?>`

agr_java_core/src/main/java/org/alliancegenome/core/config/CacheConfig.java
oblodgett: The manager object is a singleton or you could say application scoped. However it's hooked up the the hotrod client which is thread safe from my reading.

agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/DiseaseAnnotation.java
oblodgett: Eclipse does this... there is an organize imports that runs, it puts them in import order from the jar files.

agr_java_core/src/main/java/org/alliancegenome/cache/CacheService.java
cmpich: Did not know that syntax. Thanks.

agr_java_core/src/main/java/org/alliancegenome/core/config/CacheConfig.java
cmpich: I still think this is not thread-safe but we'll see.

agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/DiseaseAnnotation.java
cmpich: We should harmonize Eclipse and IntelliJ's rules for the imports.

oblodgett: Yes I agree... this will save us many headaches

PR #505, opened 2020-04-30
fix injection error

cmpich:


cmpich:
This is causing NPEs when this repository is used.

PR #507, opened 2020-05-04
more injection cleanup

cmpich:


cmpich:
Hopefully, that fixes the expression issue on BUILD.

PR #517, opened 2020-05-14
AGR-2151 add species to related genes of constucts

cmpich:


cmpich:
Just added integration tests for constructs...

PR #529, opened 2020-05-25
ES upgrade to 7.7

oblodgett:


agr_java_core/src/main/java/org/alliancegenome/core/config/ConfigHelper.java
```diff
@@ -72,7 +72,7 @@ public static void init() {
         defaults.put(ES_INDEX_SUFFIX, ""); // Prod, Dev, Stage, etc
         defaults.put(ES_DATA_INDEX, "data_index");
         defaults.put(ES_HOST, "localhost");
-        defaults.put(ES_PORT, "9300");
+        defaults.put(ES_PORT, "9200");
```
cmpich, line 75: Did the port really change?

oblodgett: We are switching from the transport client to the rest client... the transport client is getting deprecated.

PR #532, opened 2020-05-28
AGR-2072 Added VEP generated varaint names to allele index

kevinschaper:


agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleIndexerRepository.java
```diff
@@ -166,9 +166,9 @@ public AlleleDocumentCache getAlleleDocumentCache(String species) {
     }
 
     public Map<String, Set<String>> getRelatedVariants(String species) {
-        String query = "MATCH (species:Species)--(a:Allele)--(v:Variant) ";
+        String query = " MATCH (species:Species)--(a:Allele)--(v:Variant)--(tlc:TranscriptLevelConsequence) ";
```
oblodgett, line 169: With queries we have found that it is better to specify the relationship between nodes rather than using the `--` operator. What this does for instance if you start with (a:Allele) and look at ALL relationships to determine which nodes on the other end are species vs looking at only the :FROM_SPECIES relationships to determine which nodes on the other end are species.

kevinschaper: When I try this query with and without the join labels the performance doesn't seem to change in a significant way, 27ms vs 24ms to start streaming results was what I saw, and I couldn't repeat because once either has been run once they start streaming after 1ms.

PR #539, opened 2020-06-02
AGR-2174 Added construct field on alleles, fixed variant query

kevinschaper:


agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/Neo4jRepository.java
```diff
@@ -180,5 +180,21 @@ protected String addWhereClauseString(String fieldName, FieldFilter fieldFilter,
...
+
+        for (String key: keys) {
+            Set<String> values = new HashSet<>();
+            if (a.get(key) != null) { values.addAll(a.get(key)); }
+            if (b.get(key) != null) { values.addAll(b.get(key)); }
+            map.put(key, values);
+        }
+
+        return map;
+    }
```
oblodgett, line 198: @kevinschaper it would be nice to keep these "helper" functions off in another class and not in the Neo4jRepository... trying to keep separation of responsibilities. Maybe a MapUtils or QueryHelper, etc.

PR #543, opened 2020-06-04
AGR-2193 fix the cacher for gene-indexed model DAs

cmpich:


agr_java_core/src/main/java/org/alliancegenome/api/service/ModelService.java
```diff
@@ -0,0 +1,25 @@
+package org.alliancegenome.api.service;
+
+import org.alliancegenome.neo4j.entity.DiseaseAnnotation;
+import org.alliancegenome.neo4j.entity.PrimaryAnnotatedEntity;
+import org.alliancegenome.neo4j.entity.node.AffectedGenomicModel;
+import org.alliancegenome.neo4j.entity.node.GeneticEntity;
+
+public class ModelService {
```
oblodgett, line 8: When I see service I think of it in a layered system maybe this is more of a "ModelHelper"?

cmpich: A service class I typically put in some kind of business logic, thought, I agree in this case I also could use a Helper class. Will do.

PR #552, opened 2020-06-09
Cleanup of pom files

oblodgett:
Removed duplicate plugins.

### DIFF
--- a/agr_api/pom.xml
+++ b/agr_api/pom.xml
@@ -167,10 +167,13 @@
 		<finalName>agr_api</finalName>
 		<plugins>
 			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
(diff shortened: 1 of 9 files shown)


oblodgett:
@cmpich "The maven-compiler-plugin 3.6 and later versions provide a new way :"
`<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-compiler-plugin</artifactId>
    <version>3.8.0</version>
    <configuration>
        <release>9</release>
    </configuration>
</plugin>`

`<properties>
    <maven.compiler.release>9</maven.compiler.release>
</properties>`

So maybe need to clean these up before we take the branch...

oblodgett:
@cmpich ok... now we should be able to just change it in the root pom of the project and all the sub projects will change.

PR #556, opened 2020-06-11
Fix compile error

oblodgett:
@cmpich IntelliJ should catch this kind of thing?


oblodgett:
How does git let you push code then? Maybe it just sees that there is no conflict so it takes the commit.

cmpich:
Yes, I think so. Git does not compile....

PR #565, opened 2020-06-18
use weld for @Inject annotations to allow running unit tests

cmpich:


cmpich:
Not used any more. Just for sharing code purposes.

PR #581, opened 2020-06-25
AGR-2273 AGR-2279 AGR-2285 AGR-2281 AGR-2280

cmpich:


prmani:
to DiseaseRestInterface, add , schema = @Schema(type = SchemaType.STRING)) to the basedOnGene Parameter

PR #599, opened 2020-07-20
Fix speciestype AGR-2181

FB-ping:
fix species type for covid-19


oblodgett:
This branch was taken from Master... it needs to be taken from release-3.1.1

PR #610, opened 2020-07-27
AGR-2337 add human in the applicability matrix

cmpich:


oblodgett:
@cmpich this will be on hold till the new data comes through

PR #631, opened 2020-09-01
AGR-2381

cmpich:


cmpich:
I am not sure what's wrong with it. 
Also, this class is a business class that is not mapped to a Neo4j table and thus should not slow down any cypher query.

PR #637, opened 2020-09-08
AGR-2302, AGR-2398 add hgvs names to GenomicVariant object, add includeNegation in ribbon summary endpoint

cmpich:


agr_api/src/test/java/org/alliancegenome/api/tests/integration/DiseaseIT.java
```diff
@@ -706,7 +706,7 @@ public void checkDiseaseAnnotationNonDuplicated() {
     @Test
     // Test Sox9 from MGI for disease via experiment records
     public void checkDiseaseRibbonHeader() {
-        DiseaseRibbonSummary summary = diseaseService.getDiseaseRibbonSummary(List.of("MGI:98297"));
+        DiseaseRibbonSummary summary = diseaseService.getDiseaseRibbonSummary(List.of("MGI:98297"), null);
```
oblodgett, line 709: Typically its good to create another method that supports the absence of the parameter, that then calls the method with null.

oblodgett (approved):

agr_api/src/test/java/org/alliancegenome/api/tests/integration/DiseaseIT.java
cmpich: Completely, agree. Will do.

cmpich: In this case, there is no consumer for this method other than a unit test method.

PR #641, opened 2020-09-15
AGR-2370 Exclude NOT associations between diseases and genes, alleles, and models.

kevinschaper:
While I was passing through I made a lot of other relationships explicit also.


sierra-moxon (approved):

oblodgett (approved):
@kevinschaper sorry didn't see this PR come through.

PR #654, opened 2020-09-24
AGR-1712 denormalized allele rows

prmani:


agr_java_core/src/main/java/org/alliancegenome/core/translators/tdf/AlleleToTdfTranslator.java
```diff
@@ -30,24 +30,50 @@ public String getAllRows(List<Allele> annotations) {
 
     public List<AlleleDownloadRow> getAlleleDownloadRowsForGenes(List<Allele> annotations) {
 
+denormalizeAnnotations(annotations);
         return annotations.stream()
 
+                .filter(annotation -> annotation.getVariants()!=null)
+                .filter(annotation -> !CollectionUtils.isEmpty(annotation.getVariants()))
```
cmpich, line 37: Instead of 
!CollectionUtils.isEmpty(annotation.getVariants())
you can use the slightly easier-to-read version
CollectionUtils.isNotEmpty(annotation.getVariants())

agr_java_core/src/main/java/org/alliancegenome/core/translators/tdf/AlleleToTdfTranslator.java
```diff
@@ -30,24 +30,50 @@ public String getAllRows(List<Allele> annotations) {
...
                 .flatMap(Collection::stream)
                 .collect(Collectors.toList());
-
     }
 
+
+        private void denormalizeAnnotations(List<Allele> alleles) {
+            // add genetic entity info for annotations with pure genes
+
+
```
cmpich, line 52: Too many empty lines...

agr_java_core/src/main/java/org/alliancegenome/core/translators/tdf/AlleleToTdfTranslator.java
```diff
@@ -30,24 +30,50 @@ public String getAllRows(List<Allele> annotations) {
...
+
+            alleles.stream()
+                    .filter(annotation -> CollectionUtils.isEmpty(annotation.getVariants()))
+                    .forEach(annotation -> {
+                        Variant entity = createNewVariant();
```
cmpich, line 57: Why do you have to create an empty Variant object? Can you handle the null at the other end where you are expecting a non-null Variant object?

agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/Allele.java
```diff
@@ -112,5 +112,11 @@ public void setCategory(String category) {
         // this is a calculated property but the setter needs to be here
         // for deserialization purposes.
     }
+    public void addVariant(Variant entity) {
+        if (variants == null)
+            variants = new ArrayList<>();
+        if (!variants.contains(entity))
```
cmpich, line 118: WE should consider making variants a Set rather than a List to inherently have non-duplicate values.

cmpich (changes requested):
See my comments...

PR #655, opened 2020-09-24
Changed ES to use bulk Processor

oblodgett:
Multi threaded indexer caching.
Eclipse Import cleanup... @cmpich you will be happy with the imports now.


cmpich (commented):
What a  pain the import changes, making this PR affect 286 while only a few files really changed logically! Need to do synchronize IntelliJ and Eclipse import handling.

What is the speed gain by using the bulk processor? I assume it works such that you keep adding to the processor which according to the configuration decides when to persist all added documents into ES? Nice. It takes away that duty from each individual load. I wonder if the configuration works for all different loads (might depend on the size of documents, etc...).

cmpich:
Oh, wait. The changes in imports was done on purpose! Why? Is the Eclipse way of handling better than IntelliJ? Have not really looked what the difference really is.

PR #693, opened 2020-12-21
AGR-2527 Detailed Allele / Variant / Transcript table

cmpich:


agr_java_core/src/main/java/org/alliancegenome/cache/CacheAlliance.java
```diff
@@ -23,6 +24,7 @@
     //GENE_DISEASE_ANNOTATION(),
 
     ALLELE_GENE(Allele.class, (3944 * 100) + 741634), // min: 607 max: 741634 mean: 3944
+    ALLELE_VARIANT_SEQUENCE_GENE(AlleleVariantSequence.class, (3944 * 100) + 741634), // min: 607 max: 741634 mean: 3944
```
oblodgett, line 27: @cmpich Probably should run some stats on these "documents" in order to size this correctly.

agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleRepository.java
```diff
@@ -58,28 +63,75 @@ public Allele getAllele(String primaryKey) {
...
+                .stream()
+                .collect(Collectors.toMap(Transcript::getPrimaryKey, transcript -> transcript));
+        return transcriptMap;
+    }
+
+    public Set<Allele> getAllAlleles() {
+        String query = "";
+        // allele-only (no variants)
+        query += " MATCH p1=(:Species)<-[:FROM_SPECIES]-(a:Allele)-[:IS_ALLELE_OF]->(g:Gene)-[:FROM_SPECIES]-(q:Species) ";
+        query += "where not exists ((a)<-[:VARIATION]-(:Variant)) ";
```
oblodgett, line 89: @cmpich does this `(a)` need to be `(a:Allele)`?

agr_java_core/src/main/java/org/alliancegenome/neo4j/view/View.java
```diff
@@ -28,6 +28,7 @@
     public static class GeneAPI extends API { }
     public static class DiseaseAPI extends API { }
     public static class GeneAllelesAPI extends API { }
+    public static class GeneAlleleVariantSequenceAPI extends Default { }
```
oblodgett, line 31: @cmpich this should extend API not Default?

agr_java_core/src/main/java/org/alliancegenome/cache/CacheAlliance.java
cmpich: Yes, I will do this once I am sure it's working otherwise.

agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleRepository.java
cmpich: I found this syntax, though, I am sure you could also add the :Allele to it.

agr_java_core/src/main/java/org/alliancegenome/neo4j/view/View.java
cmpich: No, this is intentional at this moment. I am trying to optimize the output, i.e. reduce the payload to only the attributes really needed I am extending Default.

agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleRepository.java
oblodgett: Ya this will effect performance as in (a) would be ALL nodes connected to `[:VARIATION]`

oblodgett (approved):

PR #694, opened 2020-12-22
add crossRef -> url to payload, pheno and diseases to alleles with variants query

cmpich:


agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/Allele.java
```diff
@@ -91,7 +94,7 @@ public void setDisease(boolean hasDisease) {
         // for deserialization purposes.
     }
 
-    @JsonView({View.API.class, View.GeneAlleleVariantSequenceAPI.class})
+    @JsonView({View.API.class, View.GeneAlleleVariantSequenceAPI.class, View.GeneAlleleVariantSequenceAPI.class})
```
oblodgett, line 97: Does this need to be here twice?

cmpich: No....

oblodgett (approved):

PR #698, opened 2021-01-12
Refactored variant code into core

oblodgett:
This refactor has a couple of side effects.
1. the defaultDataSet Files now there are 2, one in the cacher with just the mods and the other in the variant indexer as it was
2. Most of the classes were moved into core and can be shared
3. I refactored the HtpCreator so the it wouldn't use as much memory and put the groupby and other code into the main thread.


oblodgett:
Also note this is a PR from variant_refactor -> variant.

cmpich (commented):
Looks good.
Except: I don't like the variant_indexer package name. It's not conventional.
Suggest: variantindexer (without underscore)

cmpich (approved):

PR #703, opened 2021-01-29
Ping test. AGR-1925

FB-ping:
make change for genetic interaction data for AGR-1925


agr_java_core/src/main/java/org/alliancegenome/es/model/query/FieldFilter.java
```diff
@@ -43,7 +47,7 @@
     HAS_DISEASE("filter.has-disease"),
     TRANSGENE_HAS_PHENOTYPE("filter.transgene-has-phenotype"),
     TRANSGENE_HAS_DISEASE("filter.transgene-has-disease"),
-    STRINGENCY("stringency"),
+    STRINGENCY("filter.stringency"),
```
oblodgett, line 50: @cmpich will this change cause issue with links from somewhere else?

cmpich: Yes, that changes the parameter for the endpoint. THe othology section might break. I am all for changing it so, we need to make that change in the UI as well.

oblodgett (approved):

cmpich (approved):
See my one comment. UI might have to change as well.

PR #704, opened 2021-02-02
AGR-2124 changes to allow excluding search faucets

adamgibs:
Backend changes for the faucet exclusion feature


oblodgett (approved):

kevinschaper (approved):
10/10, would exclude facets again

PR #708, opened 2021-02-05
set alleleRep

cmpich:


cmpich:
Forgot to set the alleleRepository for the default case.

oblodgett (approved):

PR #714, opened 2021-02-11
Added MGI file to the SPlitter

oblodgett:


cmpich (approved):
I think that is all good. Not sure what is for.

oblodgett:
Used to split a file manually.

PR #719, opened 2021-02-15
fix chromosome fixup logic

cmpich:


cmpich:
Would like to get this in quickly.

PR #731, opened 2021-02-25
AGR-2396 added species

prmani:


oblodgett (approved):

agr_java_core/src/main/java/org/alliancegenome/core/translators/tdf/AlleleToTdfTranslator.java
```diff
@@ -179,6 +180,7 @@ public String getAllTransgenicAlleleRows(List<Allele> annotations) {
 
     private TransgenicAlleleDownloadRow getBaseDownloadAlleleTransgenicRow(Allele annotation, Construct join, Publication pub) {
         TransgenicAlleleDownloadRow row = new TransgenicAlleleDownloadRow();
+        row.setSpecies(annotation.getSpecies());
```
oblodgett, line 183: @prmani you will need to call a method on the getSpecies().getMethod() there are a bunch of fields... sorry didn't catch this in the original PR.

PR #732, opened 2021-02-26
Fix interaction query

oblodgett:


oblodgett (commented):
Reverted back to the original query as the species are getting added manually. Otherwise the query takes like 8 hours.

cmpich (approved):

PR #745, opened 2021-03-08
AGR-2431 display pubs only once

prmani:


agr_java_core/src/main/java/org/alliancegenome/core/translators/tdf/AlleleToTdfTranslator.java
```diff
@@ -253,9 +253,13 @@ public String getAllVariantsRows(List<Variant> variants) {
     }
 
 
-    public List<VariantDownloadRow> getVariantDownloadRowsForAlleles(List<Variant> annotations) {
-
+    public List<VariantDownloadRow> VariantDownloadRowsForAlleles(List<Variant> annotations) {
```
oblodgett, line 256: sorry to be picky but could we use maybe `convertVariantDownloadRowsForAlleles` ?

oblodgett: or `translateVariantDownloadRowsForAlleles`

oblodgett (changes requested):

agr_java_core/src/main/java/org/alliancegenome/core/translators/tdf/AlleleToTdfTranslator.java
cmpich: I don't mind the original method name too much other than it should use 'Variant' instead of Alleles the entities passed into the method. It's a translator method and so methods are facilitating a translation from one entity to another. 
getVariantDownloadRowsForVariants.
From the signature it becomes clear that you are turning a list of variants into a list or VariantDownloadRows. that does not need to be encoded into the method name.

oblodgett:
@prmani you might need to merge this into master rather than the branch at this point.

cmpich:
The jira ticket is resolved and Prita is not working for the Alliance any longer.

PR #748, opened 2021-03-10
Fix agr 2627 orthology

FB-ping:


FB-ping:
should merge into 4.0.0, but request to merge into master

zfinannee:
Hi Ping - thank you so much for adding the Jira ticket ID to your PR titles!    We just need one tweak - the ID you use needs to be an exact match to the Jira ticket ID.  In other words,  "agr 2627" will not match -- it needs to be "AGR-2627".   Thanks!

PR #749, opened 2021-03-10
fix AGR-2627 stringencyFilter issue

FB-ping:
fix AGR-2627 orthology stringencyFilter issue


oblodgett (commented):
This may effect the data in the cache? Is this intended?

FB-ping:
> This may effect the data in the cache? Is this intended?

yes, will affect cache and reversed to original code.

FB-ping:
> This may effect the data in the cache? Is this intended?

@oblodgett more, I checked the history of those files, and reversed code and tested local cacher server.  look fine as expect.

cmpich (commented):
Looks good.

PR #753, opened 2021-03-14
more info on genes

kkarra:
added synonyms and crossrefs


oblodgett (approved):
Might want to quantify the label on `(syn)` as a Crossref if its multiple labels then its fine.

PR #756, opened 2021-03-18
genes with no chr location

kkarra:


oblodgett (approved):
Not sure about the nuance of this chance. The previous way required the genome location to be on every gene whereas the change does not require a genome location on each gene.

kkarra:
yes, there are quite a few of those..

PR #763, opened 2021-04-09
Autogenerated index

adamgibs:
Removed primary key as _id and added it to the response fields


oblodgett (changes requested):
Just delete the code... there is no point keeping legacy around in commented form.

oblodgett (approved):

PR #764, opened 2021-04-13
Variant extended

oblodgett:


adamgibs:
@oblodgett The only things I see are in SearchableItemDocument. There should be a "Stage" field (line 88 in master) and the getDocumentId method is removed in master but commented out here. Although both may be fixed when the conflicts are resolved.

oblodgett:
@adamgibs ya... I guess what I don't understand is why github is not showing that difference, with the stage field. We did make that change to the documentId field in this branch... and I would agree that we need to get rid of the code but that logic I think needs to go into a different method.... maybe it's not an issue?

adamgibs:
@oblodgett I think the only reason that logic is in there is because the _id has a character limit, but if they're all auto generated it not be needed at this point, right? Unless we just want to limit HGVS name length for display purposes.

oblodgett:
@adamgibs Oh ok... if that was the reason... I will just remove the code.

cmpich (approved):

oblodgett:
@adamgibs for some reason I overlooked the conflicting files... in there was the missing stage map issue...

adamgibs (approved):

adamgibs:
@oblodgett Got it. That makes

PR #767, opened 2021-04-16
Fixed != == bug

oblodgett:


adamgibs (commented):
@oblodgett Just so I understand the logic (using crossReferences as an example):

- 	The first null check checks if the cossReferences HashSet has any items associated with the document's primaryKey
- 	If so, Then the second check looks to see if crossReferences have been set on the document
- 		If not, an empty HashSet is initialized
- 	Finally, whatever crossReferences are part of the IndexerCache, are appended to the whatever crossReferences are in the documents

		
Is that correct?

oblodgett:
@adamgibs ya... so I what I am trying to fix is the:
`document.getChromosomes().addAll(chromosomes.get(id));`
Where `document.getChromosomes()` is null... now why is it null?
I changed the line in the SearchItemDocument from
`Set<String> chromosomes = new HashSet<String>();`
to
`Set<String> chromosomes;`
Now you might ask why remove that?
Well in the human variant indexer when it has to initialize 600M of these documents not initializing an empty set is quickest then having all these empty sets hanging around that are not used.

adamgibs (approved):
@oblodgett Got it. Only initializing the HashSets when they are needed makes sense to me.

PR #768, opened 2021-04-19
Fix searchable item indexer performance

oblodgett:


cmpich (approved):

oblodgett:
@cmpich one change that is coming through with this PR is that a Repo has to be closed. So `repo = new GeneRepository();` at some point later a `repo.close();` will need to be called. Otherwise the program will not exit. Before we were using a single static Neo session which meant we were not multithreading into neo. With this change we get much better performance out of neo. Indexer run time is like 2 minutes minus a single query that takes about 10. Cacher I was about to get down to a 20 minute runtime 15 of those minutes are a single Allele query. I don't know what this is going to do to the API. Maybe the solution is to not have a repo defined at the class level in the services, but to open and close the repo in each method call. This might be a slight bit slower, but it would be multi threaded across the API. So one query wouldn't take the API down as we have seen in the past.

PR #778, opened 2021-04-24
use regex for *RNA_gene, deboost pseudogenes to compensate for human boost

gildossantos:
For genes, human pseudogenes should rank below MOD non-pseudogenes, so "boosting" by 0.5X to counteract the 1.5X human boost.
Also, I'm suggesting use of `.*RNA` regex, which should catch almost all genes having biotype in the ncRNA_gene branch; I'm assuming this is possible.


cmpich (approved):
Looks good.

oblodgett (approved):

adamgibs (approved):

PR #782, opened 2021-04-27
AGR-2845

adamgibs:
Added View.AlleleVariantSequenceConverterForES.class JsonView annotations over getSynonymList(), crossReferences, and getSecondaryIdsList() in GeneticEntity.


oblodgett:
@adamgibs so CrossReferences are a little complicated... If you notice on line 122 there is an "override" `@JsonProperty(value = "crossReferences")` which means even though the mapper will call that getter because the json property the getter is sorta remapped to the other method. So you want to the put the view with the `@JsonProperty(value = "crossReferences")`

adamgibs:
@oblodgett Okay, I've pushed out an update.

oblodgett (approved):

PR #786, opened 2021-04-30
AGR-2565 get allele info from ES for LTP and HTP

cmpich:


agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleVariantSequenceConverter.java
```diff
@@ -64,8 +65,13 @@
...
             variant.setSpecies(species);
-            variant.setStart(String.valueOf(ctx.getStart()));
-            variant.setEnd(String.valueOf(endPos));
+            GenomeLocation location = new GenomeLocation();
+            location.setStart((long) ctx.getStart());
+            location.setEnd((long) ctx.getEnd());
+            Chromosome chromosome = new Chromosome();
+            chromosome.setPrimaryKey(ctx.getChr());
+            location.setChromosome(chromosome);
+            variant.setLocation(location);
```
oblodgett, line 74: We need to test the performance of this change on the human indexer.

oblodgett (approved):

PR #788, opened 2021-05-04
AGR-2842 added species to TranscriptLevelConsequence

adamgibs:
After we changed line 114 in the AlleleVariantSequenceConverter to use getNameKey(), it still didn't appear to append the species abbreviation to the gene symbol on build. So I tried running the variant indexer with the debugger and noticed that species was always null within the getNameKey() method. 
	
In TranscriptLevelConsequence, setSpecies was never called on associatedGene and that the infos doesn't  contain species. So I passed down species from AlleleVariantSequenceConverter to TranscriptLevelConsequence

### DIFF
--- a/agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleVariantSequenceConverter.java
+++ b/agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleVariantSequenceConverter.java
@@ -99,6 +99,7 @@ public class AlleleVariantSequenceConverter {
             avsDoc.setVariant(variant);
             if (htpConsequences != null) {
                 for (TranscriptLevelConsequence c : htpConsequences) {
+                    c.getAssociatedGene().setSpecies(species);
                     if(!transcriptsProcessed.contains(c.getTranscriptID())) {
                         transcriptsProcessed.add(c.getTranscriptID());
                         if(first) {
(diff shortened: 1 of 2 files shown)


oblodgett:
I would do this differently so that there wasn't such a breaking change. Right after line 116 in the AVS Converter I would just add the following:
`c.getAssociatedGene().setSpecies(species)` this way the species object doesn't have to be passed around its already a class level variable.

adamgibs:
@oblodgett Update has been pushed out

oblodgett:
I think you might want it to happen outside the larger if statement: `if(!transcriptsProcessed.contains(c.getTranscriptID())) {` ??

adamgibs:
@oblodgett Okay, sure. I will do that. From what I can tell, it just needs to happen before the `genes.add(c.getAssociatedGene().getNameKey());` line so that species isn't null within the getNameKey() method:         
```
if (species != null) {
            nameKey += " (" + species.getType().getAbbreviation() + ")";
        }
```
So should I add the change prior to the `if(!transcriptsProcessed.contains(c.getTranscriptID()))`, probably on line 101?

oblodgett:
Ya... before (line 102) or after (line 119) the if statement I don't think really matters. It just needs to be outside of the if statement, so that it it happens for all TLC's that are created.

oblodgett (approved):

PR #796, opened 2021-05-13
AGR-2892

adamgibs:


oblodgett (commented):
Are these going to get changed via the Human File? If so we need to wait on the new file before making code changes?

adamgibs:
Oh, that's a good point. I'm not sure. I'll hold off on these changes then.

gildossantos:
I'm still seeing the odd variant types that this PR was intended to suppress. Has the new human file had enough time to percolate through to the search UI on BUILD? If so, then should we reopen this PR now?

adamgibs:
@gildossantos I'm not sure about this. @oblodgett Is the new human file on build yet?

oblodgett:
No we are trying to trace down a major slowdown in the code that is causing the indexer to go from 1.25 hours to 4.5 hours. Once we can sort that out... will implement the new human file.

gildossantos:
The human variants are back on BUILD, which I take to mean that the new human VCF file has been processed and all changes have percolated to the front end. If that's true, then the new file didn't fix the issue with the odd variant types, so perhaps we should reopen this PR now?

oblodgett:
Nope... its just that the human indexer happened to run and import an new index but the files are being uploaded right now.

PR #798, opened 2021-05-14
AGR-2845 added gene synonyms and cross references to htp variants

adamgibs:
This seems to work on my local machine, but I'm sure this is not the optimal way to do it. Feedback is definitely welcome. 

I instantiated `GeneDocumentCache` in `HtpVariantCreation` in order to avoid compilation errors. However, it's unclear to me what `HtpVariantCreation` is doing.

Added a `crossReferencesList` to `GeneticEntity` in order to have something similar to the synonyms list. I added its get method in `Gene` because I was running into a conflict with the `getCrossReferencesMap`  getter.

### DIFF
--- a/agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/Gene.java
+++ b/agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/Gene.java
@@ -149,17 +149,4 @@ public class Gene extends GeneticEntity implements Comparable<Gene> {
     }
 
 
-    @JsonView(value = {View.AlleleVariantSequenceConverterForES.class})
-    @JsonProperty(value = "crossReferences")
-    public List<String> getCrossReferencesList() {
-        if (crossReferences != null) {
-            List<String> list = new ArrayList<>();
-            for (CrossReference crossReference : crossReferences) {
-                list.add(crossReference.getName());
-            }
-            return list;
-        } else {
-            return new ArrayList<>();
-        }
-    }
 }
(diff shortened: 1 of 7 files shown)


agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/Gene.java
```diff
@@ -146,4 +147,19 @@ public boolean equals(Object o) {
     public int hashCode() {
         return Objects.hash(primaryKey);
     }
+
+
+    @JsonView(value = {View.AlleleVariantSequenceConverterForES.class})
+    @JsonProperty(value = "crossReferences")
```
cmpich, line 153: There is a method in the parent class, GeneticEntity, that has a JsonPropery annotated like this. I am not sure why you need a new method. Does the old one on GeneticEntity not suffice?

oblodgett: Adam and I went over a change to this, that will be in the GeneticEntity class... basically 3 methods: `getCrossReferences` which will return the list of neo objects: `getCrossReferenceMap` which will return what the UI is looking for and `getCrossReferenceList` which will return a list of strings of the crossreference "name" field, and is what the search needs. All the JsonProperties go away and we go back to the method names. The UI will need to change to use `crossReferenceMap` in the json but Adam will be putting in a change to that in the UI repo.

adamgibs:
@oblodgett An update has been pushed out.

oblodgett (approved):
Just hold on to bringing this PR in, until the UI can be changed, good work thanks.

agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/Gene.java
cmpich: Sounds good.

adamgibs:
@oblodgett I've resolved the conflicts. Requesting review just to make sure overwrite one of your changes.

oblodgett:
Ya everything looks good... lets bring it in.

PR #799, opened 2021-05-17
AGR-2863 Variant detail page: create ES retrieval logic for transcript table

cmpich:


agr_java_core/src/main/java/org/alliancegenome/es/index/VariantESDAO.java
```diff
@@ -25,14 +24,18 @@
...
+import java.util.Map;
 
 import static java.util.stream.Collectors.toList;
 import static org.elasticsearch.index.query.QueryBuilders.boolQuery;
 
+@Log4j2
 public class VariantESDAO {
 
-    private Log log = LogFactory.getLog(getClass());
+    public static final String SITE_INDEX = "site_index";
```
oblodgett, line 38: @cmpich SITE_INDEX needs to be pulled from the `ConfigHelper.getEsIndex()`

oblodgett (changes requested):

PR #800, opened 2021-05-19
add genomeLocation to Gene on variant, add sorting of variants

cmpich:


agr_api/src/main/java/org/alliancegenome/api/controller/VariantController.java
```diff
@@ -24,12 +25,17 @@
...
 
+    private VariantRepository variantRepository = new VariantRepository();
+
     @Inject
     private HttpServletRequest request;
 
     @Override
     public Variant getVariant(String id) {
-        Variant variant = variantService.getVariantById(id);
+        Variant variant = variantRepository.getVariant(id);
```
oblodgett, line 35: So we try to maintain a layered architecture. I know it seems that there are maybe redundant layers but they do hav ea purpose. the Service layer is for business logic.

oblodgett: Controller -> Service -> Repo / DAO this and ideally the logic that is here in the control should be moved to the service. And the only code in the controller should be for handling errors.

oblodgett (changes requested):

agr_api/src/main/java/org/alliancegenome/api/controller/VariantController.java
cmpich: Fully agree. Refactored that logic into the service class.

PR #801, opened 2021-05-24
updated site_indexer added allele objects

jt15:


agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/relationship/GenomeLocation.java
```diff
@@ -25,6 +25,7 @@ public String getChromosomeName() {
         return chromosome.getPrimaryKey();
     }
 
+    @JsonView({View.AlleleVariantSequenceConverterForES.class})
```
oblodgett, line 28: JSonView doesn't go on a setMethod

agr_java_core/src/main/java/org/alliancegenome/es/index/site/cache/IndexerCache.java
```diff
@@ -142,5 +144,110 @@ public void addCachedFields(SearchableItemDocument document) {
 
 
     }
+    public void addCachedAlleleFields(Iterable<AlleleVariantSequence> documents) {
+        for (AlleleVariantSequence document : documents) {
+            addCachedAlleleFields(document);
+        }
+    }
+    public void addCachedAlleleFields(AlleleVariantSequence document) {
```
oblodgett, line 152: This needs to not be a copy and paste of `addCachedFields`

jt15: That's true, but this would be quick fix

agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/relationship/GenomeLocation.java
jt15: yes fixed it now

oblodgett (approved):

PR #802, opened 2021-05-24
Variant Performance and code cleanup

oblodgett:


oblodgett:
@jt15 this adds your PR into the mix and this should get us back to the 200k records per second.

cmpich (approved):
Looks good.

adamgibs (approved):

PR #813, opened 2021-05-28
Added release info API endpoints AGR-2901

oblodgett:


oblodgett:
For some reason the dates are not working from Neo @cmpich do you know how to fix this?

PR #826, opened 2021-06-08
Added fix for empty hgvsNomenclature

oblodgett:


cmpich (approved):
What if the hgvsNomenclature does not have a colon? The method with throw a NPE then...
If it is either empty or has colon then this will work, maybe a comment would be good to capture this.

PR #828, opened 2021-06-08
Htp index

jt15:


agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/GeneticEntity.java
```diff
@@ -137,14 +137,14 @@ public void setSecondaryIdsList(List<String> secondaryIdsList) {
         }
         return crossReferencesMap;
     }
-
+    @JsonProperty(value = "crossReferences")
```
oblodgett, line 140: We need to not have crossReferences point to the Map it needs to point to the list of Neo objects. This was on purpose by Adam's PR from a couple of days ago.

oblodgett (changes requested):

agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/GeneticEntity.java
oblodgett: if it needs to be included / excluded we need to use a View for that. It also depends on what we want to put into the ES index. But we have three items:
crossReferences -> list of neo objects
crossReferencesMap -> this map method
crossReferencesList -> list of crossReference names

oblodgett (approved):

PR #829, opened 2021-06-08
Indexer snapshot timeout fix

oblodgett:
@cmpich maybe there is a better way to do this with "Futures"?


agr_java_core/src/main/java/org/alliancegenome/es/util/IndexManager.java
```diff
@@ -293,16 +309,22 @@ public void createSnapShot(String repo, String snapShotName, List<String> indice
...
             request.snapshot(snapShotName);
             request.indices(indices);
             request.waitForCompletion(true);
 
-            getClient().snapshot().create(request, RequestOptions.DEFAULT);
+            getClient().snapshot().createAsync(request, RequestOptions.DEFAULT, requestListener);
+            
+            while(!requestFinished) {
+                TimeUnit.SECONDS.sleep(5);
+            }
```
oblodgett, line 325: The goal here is that this method HAS to wait until it's finished. However `getClient().snapshot().create(request, RequestOptions.DEFAULT);` times out at 1800 seconds.

adamgibs (approved):

PR #834, opened 2021-06-09
move transcriptLevelConsequence attribute over from String to List<St…

cmpich:
…ring>


oblodgett:
@cmpich can we just make the change instread of renaming the column?

PR #838, opened 2021-06-09
AGR-2845 added geneSynonyms and geneCrossReferences to top level of v…

adamgibs:
…ariant doc

### DIFF
--- a/agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleVariantSequenceConverter.java
+++ b/agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleVariantSequenceConverter.java
@@ -78,6 +78,7 @@ public class AlleleVariantSequenceConverter {
 
             variant.setNucleotideChange(vcfAllele.getBaseString()); // variantDocument.setVarNuc(a.getBaseString());
             boolean first = true;
+            boolean firstConsequence = true;
             Set<String> molecularConsequences = new HashSet<>();
             Set<String> genes = new HashSet<>();
             Set<String> geneIds = new HashSet<>();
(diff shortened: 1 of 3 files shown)


agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleVariantSequenceConverter.java
```diff
@@ -112,27 +113,24 @@
...
-                            consequenceGene.setCrossReferencesList(crossReferencesList);
-                        }
-                        
-                    }
 
                     if(consequenceGene != null) {
                         consequenceGene.setSpecies(species);
                     }
+
+                    if(firstConsequence){
```
adamgibs, line 126: I created a second first check because I wasn't sure if it would make sense to have this logic under the `if(!transcriptsProcessed.contains(transcriptID))` check

oblodgett: Why not just `break;` inside the if statement once the synonym and crossref's are set?

adamgibs:
This change would remove gene synonyms and crossReference from the associated gene for both variant objects and TLCs. If those fields are needed for the pages, then I can definitely revert that part of the change.

agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleVariantSequenceConverter.java
oblodgett: Oh never mind the rest of the loop needs to run for each CSQ

adamgibs: Yeah, I still feel like there's a better way than what I have.

oblodgett: You might want to put the `firstConsequence = false;` inside the if statement so they will be set.

adamgibs: So put it within the `if(geneCache != null && consequenceGene != null)` check? Would that be in case that statement evaluates to true on an iteration other than the first, but not on the first iteration?

oblodgett (approved):

PR #840, opened 2021-06-11
Variant index mappings

oblodgett:


agr_java_core/src/main/java/org/alliancegenome/es/index/site/schema/settings/VariantIndexSettings.java
```diff
@@ -26,11 +26,11 @@ public void buildSettings() throws IOException {
                 //.field("merge.scheduler.max_thread_count", "10")
                 //.field("merge.policy.floor_segment", "200mb")
                 //.field("merge.policy.max_merged_segment", "500mb")
-                .field("translog.sync_interval", "15s")
+                .field("translog.sync_interval", "60s")
                 .field("translog.flush_threshold_size", "4096mb")
                 .field("translog.durability", "async");
                 //.field("codec", "best_compression");
-            buildAnalysis(true);
+            //buildAnalysis(true);
```
oblodgett, line 33: @jt15 we may need to bring back the `buildAnalysis` for the variant indexes but for now going to try without it.

agr_variant_indexer/src/main/java/org/alliancegenome/indexer/variant/es/managers/SourceDocumentCreation.java
```diff
@@ -433,7 +433,7 @@ public void run() {
         private final ObjectMapper mapper = new ObjectMapper();
 
         public void run() {
-            mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
+            //mapper.setSerializationInclusion(JsonInclude.Include.NON_NULL);
```
cmpich, line 436: Why are you commenting this out?

oblodgett: `        /**
         * Value that indicates that only properties with non-null
         * values are to be included.
         */
        NON_NULL,`
        
`         * With 2.7, definition has been tightened back to only containing types explained
         * above (null, absent, empty String, empty containers), and now
         * extended definition may be specified using {@link #NON_DEFAULT}.
         */
        NON_EMPTY,`

oblodgett: you don't need both since we have NON_EMPTY we don't need NON_NULL. And I wanted to make sure they were not conflicting with each other.

cmpich (approved):

PR #841, opened 2021-06-11
AGR-2892 suppressed 'MIXED' and 'SYMBOLIC' variant type facets in hum…

adamgibs:
…an variants

### DIFF
--- a/agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleVariantSequenceConverter.java
+++ b/agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleVariantSequenceConverter.java
@@ -32,7 +32,7 @@ public class AlleleVariantSequenceConverter {
 
         SOTerm variantType = new SOTerm();
 
-        if(!"SYMBOLIC".equalsIgnoreCase(ctx.getType().name()) && !"MIXED".equalsIgnoreCase(ctx.getType().name())){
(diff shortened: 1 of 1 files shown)


agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleVariantSequenceConverter.java
```diff
@@ -31,27 +31,31 @@
         }
 
         SOTerm variantType = new SOTerm();
-        variantType.setName(ctx.getType().name().toUpperCase());
-        variantType.setPrimaryKey(ctx.getType().name());
-        if ("INDEL".equals(ctx.getType().name())) {
-            variantType.setName("delins");
-            variantType.setPrimaryKey("delins");
+
+        if(!"SYMBOLIC".equalsIgnoreCase(ctx.getType().name()) && !"MIXED".equalsIgnoreCase(ctx.getType().name())){
```
oblodgett, line 35: Does this have to do a `equalsIgnoreCase` or can we use a straight `equals` ?

adamgibs: I'm sure equals would work. I probably did that as an extra safeguard just in case.

oblodgett (approved):

PR #852, opened 2021-06-15
AGR-3018 added prefix 'ZFIN:' to geneCache lookup for zfin files

adamgibs:


agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/TranscriptLevelConsequence.java
```diff
@@ -189,8 +189,13 @@ public TranscriptLevelConsequence(String[] header, String[] infos, GeneDocumentC
 
             molecularConsequences = Arrays.asList(infos[1].split("&"));
             impact = infos[2];
-            
-            associatedGene = geneCache.getGeneMap().get(infos[4]);
+
+            if(header[37].substring(0, 4).equals("ZFIN")){
```
oblodgett, line 193: @adamgibs is this a bug in the VCF file?

adamgibs: @oblodgett I don't know if it's a bug or supposed to be that way, but it causes zfin docs to not get geneSynonyms or geneCrossReferences for search

adamgibs:
Per @oblodgett the fix should occur in the VCF file.

agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/TranscriptLevelConsequence.java
oblodgett: Lets check with Mark QT before closing this PR.... this might be our only options for getting it in for release.

gildossantos: I noticed an issue with ZFIN gene IDs in the GFF file - they lack the expected "ZFIN:" prefix, which coincidentally is missing in the fish variant links to gene pages. See AGR-3020. Is it possible this is the root cause?

adamgibs: @gildossantos That would make sense.

oblodgett: @adamgibs @gildossantos looks like the GFF change for ZFIN has rolled through. I don't think we need this code anymore.

adamgibs: @oblodgett Sounds good. I will close this request then

PR #853, opened 2021-06-15
Allele variants

jt15:


agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleVariantSequenceConverter.java
```diff
@@ -118,7 +118,7 @@
                 }
 
             }
-
+            agrAllele.setPrimaryKey(variant.getPrimaryKey());
```
oblodgett, line 121: @jt15  shouldn't this be done in the index? This is an issue in other places.

jt15: sure we can do it in the api

oblodgett: OH... I got mixed up... yes this is where this should be done... and you shouldn't need all that code in the AlleleVariantIndexService.java if the AlleleVariantSequenceConverter.java makes sure there is a primaryKey before indexing the allelel.

oblodgett: Unless the LTP doesn't have one?

jt15: do want id field to put it back in the AlleleVariantSequenceConverter.java and remove the lines from api code?

oblodgett: ya sorry... its been a long day I got mixed up... if we get the primaryKey into the index than it will take care of multiple issue elsewhere

jt15: yes

jt15: its reverted now. added id to avsconverter and removed api code lines

oblodgett (approved):

PR #854, opened 2021-06-15
Fix for Allele API

oblodgett:


agr_java_core/src/main/java/org/alliancegenome/api/service/AlleleColumnFieldMapping.java
```diff
@@ -31,7 +31,9 @@ public AlleleColumnFieldMapping() {
         mapColumnAttribute.put(GENE_ALLELE_CATEGORY, entity -> Set.of(entity.getCategory()));
         mapColumnAttribute.put(GENE_ALLELE_VARIANT_TYPE, entity -> {
             if (entity.getVariants() != null) {
-                return entity.getVariants().stream().map(variant -> variant.getVariantType()!=null?variant.getVariantType().getName():"").collect(toSet());
```
oblodgett, line 34: @jt15 I guess we both fixed this.

jt15: yes, how long do the updates get reflects on the build site?

PR #866, opened 2021-06-21
Allele sort filter

jt15:


agr_api/src/main/java/org/alliancegenome/api/rest/interfaces/GeneRESTInterface.java
```diff
@@ -82,8 +82,8 @@ Gene getGene(
             @QueryParam("filter.synonyms") String synonym,
             @Parameter(in = ParameterIn.QUERY, name = "filter.variantType", description = "Variant types", schema = @Schema(type = SchemaType.STRING))
             @QueryParam("filter.variantType") String variantType,
-            @Parameter(in = ParameterIn.QUERY, name = "filter.variantConsequence", description = "Consequence", schema = @Schema(type = SchemaType.STRING))
-            @QueryParam("filter.variantConsequence") String consequence,
+            @Parameter(in = ParameterIn.QUERY, name = "filter.molecularConsequence", description = "Consequence", schema = @Schema(type = SchemaType.STRING))
+            @QueryParam("filter.molecularConsequence") String consequence,
```
oblodgett, line 86: @jt15 is the UI excepting this change? or is this to fix an existing issue?

jt15: I don't know. but there is no filter.molecularConsequence parameter in the API method that's why molecular consequence filter is not working. When changed variantConsequence to molecularConsequence the filter is working. May be we can wait for @cmpich 's input on this.

oblodgett: Ok

oblodgett (approved):

PR #868, opened 2021-06-21
Allele indexer refactored

jt15:
@adamgibs the allele indexer is refactored to put a fix for indexing associated gene which is being used in the allele variant detail table, as well to also index transgenic alleles. See if it is working as expected and not missing any transgenic alleles, an issue  AGR-3023 which was recently resolved.

### DIFF
--- a/agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleRepository.java
+++ b/agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleRepository.java
@@ -440,7 +440,7 @@ public class AlleleRepository extends Neo4jRepository<Allele> {
     }
     public Map<String, Allele> getAllAlleleVariants() {
 
-        String query = "MATCH p1=(:Species)-[:FROM_SPECIES]-(a:Allele)";
+        String query = "MATCH p1=(:Species)-[:FROM_SPECIES]-(a:Allele)--(g:Gene)";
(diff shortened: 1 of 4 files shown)


adamgibs (approved):

agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleRepository.java
```diff
@@ -440,7 +440,7 @@ private String getCypherQuery(String relationship) {
     }
     public Map<String, Allele> getAllAlleleVariants() {
 
-        String query = "MATCH p1=(:Species)-[:FROM_SPECIES]-(a:Allele)";
+        String query = "MATCH p1=(:Species)-[:FROM_SPECIES]-(a:Allele)--(g:Gene)";
```
oblodgett, line 443: I don't know if this is a long running query, but we have found its better to specify the relationship between the nodes to speed up the query so in this case I think it should be a `(a:Allele)-[:IS_ALLELE_OF]-(g:Gene)`

oblodgett (approved):

agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleRepository.java
jt15: @oblodgett fixed neo query as suggested.

PR #872, opened 2021-06-23
Added code to clean up snapshots from ES repos

oblodgett:


agr_java_core/src/main/java/org/alliancegenome/es/util/IndexManager.java
```diff
@@ -454,6 +454,25 @@ public void listRepo(String repo) { // ES Util
             log.info(info.snapshotId() + "[" + print + "] " + end);
         }
     }
+    
+
+    public void cleanSnapShots(String repo, String snapShotName) {
+        List<SnapshotInfo> list = getSnapshots(repo);
+        TreeMap<Date, SnapshotInfo> map = new TreeMap<>();
```
cmpich, line 461: Why are you using a TreeMap, you are not making any use of sorting. HashMap should be sufficient.

agr_java_core/src/main/java/org/alliancegenome/es/util/IndexManager.java
```diff
@@ -454,6 +454,25 @@ public void listRepo(String repo) { // ES Util
...
+        for(SnapshotInfo info: list) {
+            String[] array = info.snapshotId().getName().split("_");
+            Date d = new Date(Long.parseLong(array[array.length - 1]));
+            if(info.snapshotId().getName().contains(snapShotName)) {
+                map.put(d, info);
+            }
+        }
+        if(map.size() > 0) {
+            map.remove(map.lastKey());
```
oblodgett, line 470: @cmpich removing the last key as in the newest so keeping the newest snapshot

PR #897, opened 2021-07-26
Variant

oblodgett:


cmpich (approved):
Looks good. I assume those are my changes except the ES change.

oblodgett:
@cmpich @jt15 and I have been looking at this and something doesn't seem right.

cmpich:
What is wrong?

cmpich:
I hope to be able to take another closer look until Monday.

PR #945, opened 2022-03-04
Fix urls to be the same field name as in neo

oblodgett:


oblodgett:
@cmpich this is going to require a UI change will work on getting that in as well.

PR #953, opened 2022-09-12
SCRUM-1968 fix orthologous record order: use phylogenetic ordering

cmpich:


agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/node/Species.java
```diff
@@ -30,7 +30,7 @@ public class Species extends Neo4jEntity implements Comparable<Species> {
     @JsonView({View.Default.class,View.AlleleVariantSequenceConverterForES.class})
     private String dataProviderShortName;
     
-    @JsonView({View.DiseaseCacher.class, View.OrthologyCacher.class})
+    @JsonView({View.DiseaseCacher.class, View.Orthology.class})
```
oblodgett, line 33: Are you sure you can make this change? Is there other places that this view is used?

oblodgett (approved):

PR #964, opened 2022-09-20
Removed chromosome from repo queries

oblodgett:


oblodgett:
This is linked to: https://github.com/alliance-genome/agr_loader/pull/636

oblodgett:
PS the Chromosome is on the GenomicLocation object so this extra connection is redundant, and this should speed up the Neo4j queries.

agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/VariantRepository.java
```diff
@@ -73,16 +73,4 @@ public List<Allele> getAllelesOfVariant(String variantID) {
 				.collect(Collectors.toList());
 	}
 
-	public GenomeLocation getGenomeLocation(String geneID) {
```
cmpich, line 76: Where was this method being called? I don't see a removal of a calling unit. Maybe it was never used?

oblodgett: It wasn't being used.

cmpich: Great. Then it should go out.

cmpich (approved):
Looks good.

PR #968, opened 2022-10-07
SCRUM-2013 disease annotation indexer

cmpich:


agr_indexer/src/main/java/org/alliancegenome/indexer/DiseaseAnnotationRESTInterface.java
```diff
@@ -0,0 +1,31 @@
...
+import org.alliancegenome.curation_api.model.entities.AGMDiseaseAnnotation;
+import org.alliancegenome.curation_api.model.entities.AlleleDiseaseAnnotation;
+import org.alliancegenome.curation_api.model.entities.GeneDiseaseAnnotation;
+import org.alliancegenome.curation_api.response.SearchResponse;
+
+import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
+
+@Path("/api")
+public interface DiseaseAnnotationRESTInterface {
```
oblodgett, line 12: @cmpich we should be able to use the interfaces from the library. And you can configure the auth headed in the configClient vs on every call.

agr_indexer/src/main/java/org/alliancegenome/indexer/RestConfig.java
```diff
@@ -0,0 +1,41 @@
...
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import org.alliancegenome.api.json.APIBeanSerializerModifier;
+import org.alliancegenome.core.config.ConfigHelper;
+import si.mazi.rescu.ClientConfig;
+import si.mazi.rescu.serialization.jackson.DefaultJacksonObjectMapperFactory;
+import si.mazi.rescu.serialization.jackson.JacksonObjectMapperFactory;
+
+import javax.ws.rs.HeaderParam;
+
+public class RestConfig {
```
oblodgett, line 15: @cmpich you should be able to use the ObjectMapper from the RestDefaultObjectMapper in the curation.config class so that we don't need to maintain both.

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/linkml/DiseaseAnnotationIndexer.java
```diff
@@ -0,0 +1,51 @@
...
+import org.alliancegenome.indexer.DiseaseAnnotationRESTInterface;
+import org.alliancegenome.indexer.RestConfig;
+import org.jboss.logging.Logger;
+import si.mazi.rescu.RestProxyFactory;
+
+
+public class DiseaseAnnotationIndexer {
+
+    private Logger log = Logger.getLogger(getClass());
+    private DiseaseAnnotationRESTInterface api = RestProxyFactory.createProxy(DiseaseAnnotationRESTInterface.class, ConfigHelper.getCurationApiBaseUrl(), RestConfig.config);
```
oblodgett, line 17: Switch out the RestConfig.config for one that adds the auth header or not depending if its set.

oblodgett (changes requested):

PR #1001, opened 2023-02-07
SCRUM-2019 add parent IDs for each disease term

cmpich:


agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
```diff
@@ -123,6 +116,17 @@ private List<GeneDiseaseAnnotationDocument> createGeneDiseaseAnnotationDocuments
 		return ret;
 	}
 
+	private static Map<String, Set<String>> closureMap = null;
+
+	// Get Closure for disease ontology
+	private static Map<String, Set<String>> getClosure() {
+		if (closureMap == null) {
+			DiseaseRepository diseaseRepository = new DiseaseRepository();
```
oblodgett, line 124: This DiseaseRepository should be opened as part of the main class and not on each call to getClosure.

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
```diff
@@ -123,6 +116,17 @@ private List<GeneDiseaseAnnotationDocument> createGeneDiseaseAnnotationDocuments
 		return ret;
 	}
 
+	private static Map<String, Set<String>> closureMap = null;
```
oblodgett, line 119: These variables should be defined at the top of the class with the others.

agr_java_core/src/main/java/org/alliancegenome/api/entity/DiseaseAnnotationDocument.java
```diff
@@ -27,6 +27,7 @@ public class DiseaseAnnotationDocument extends SearchableItemDocument {
 	private Reference singleReference;
 	private Set<Reference> references;
 
+	Set<String> parentSlimIDs;
```
oblodgett, line 30: should be private?

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
cmpich: This method is only executed once until the map is initialized. Moved it anyway.

agr_java_core/src/main/java/org/alliancegenome/api/entity/DiseaseAnnotationDocument.java
cmpich: Yes.

oblodgett (approved):

PR #1016, opened 2023-02-15
Scrum 2551

oblodgett:


oblodgett:
@cmpich @chris-grove there is no magic... however this one liner change took 3 days to find. Once the index hits this field limit it does not allow any new documents with "new" fields. Hence after adding all the diseaseAnnotation documents there were no fields left for the allele documents.

adamgibs (approved):

PR #1019, opened 2023-02-17
Updated default tag for linux env.

christabone:


oblodgett (approved):
Sorry didnt catch all these

mluypaert:
@christabone @oblodgett  Why changing the defaults rather than using env variables when running the Docker build's? We originally set these `latest` as default tag value to prevent accidental overwrites of GoCD-generated images like `stage`, or production images like specific release numbers.

christabone:
I'll defer to Olin on this one, this is his decision that I was implementing in the code.

christabone:
The latest images are also gone, so setting them as default doesn't really make much sense anymore from that perspective.

PR #1020, opened 2023-02-18
SCRUM-2409 filter on negated flag

cmpich:


abecerra (approved):

oblodgett (approved):
Typically big B booleans are used for input on endpoints

PR #1021, opened 2023-02-22
SCRUM-2409 SCRUM-2374 use diseaseRelationNegation field for filtering and creating the full value set

cmpich:


agr_api/src/main/java/org/alliancegenome/api/service/DiseaseESService.java
```diff
@@ -100,6 +100,12 @@ public JsonResultResponse<GeneDiseaseAnnotationDocument> getRibbonDiseaseAnnotat
 					String[] elements = filterValue.split("\\|");
 					Arrays.stream(elements).forEach(element -> orClause.should(QueryBuilders.termQuery(filterName, element)));
 					bool.must(orClause);
+				} else if (filterValue.contains(" ")) {
```
oblodgett, line 103: @cmpich im not sure this is a good check... I think we can handle filtering better.

cmpich: I need to take each individual word and check for any contains matches. How can we do better?

oblodgett: Use a simple query and the "AND" operator

oblodgett: Or couldn't you just change the spaces to underscores and then you would have an exact match?

cmpich: I don't want an exact match. 
<query1> <query2> <query3>
should be interpreted as
<*query1*>AND<*query2*>AND<*query3*>
ie as wildcard matches for each query string connected with AND.

oblodgett: Which field is this for?

cmpich: Disease name

PR #1023, opened 2023-02-24
Added changes

oblodgett:


agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
```diff
@@ -106,9 +106,12 @@ private List<GeneDiseaseAnnotationDocument> createGeneDiseaseAnnotationDocuments
 
 				if (gdad == null) {
 					gdad = new GeneDiseaseAnnotationDocument();
-					
 					gdad.setSubject(entry.getValue().getLeft());
-					int order = SpeciesType.getTypeByID(entry.getValue().getLeft().getTaxon().getCurie()).getOrderID();
+					SpeciesType type = SpeciesType.getTypeByID(entry.getValue().getLeft().getTaxon().getCurie());
+					int order = 10;
```
cmpich, line 111: I would take a higher number to always put a record to the end that has no taxon order number.

oblodgett: Ok they are set to 100

cmpich (approved):

PR #1031, opened 2023-03-09
SCRU-2684 honor the include-negated-annotation flag

cmpich:


agr_api/src/main/java/org/alliancegenome/api/controller/GeneController.java
```diff
@@ -612,7 +612,7 @@ public ExpressionSummary getExpressionSummary(String id) {
 	// the List passed in here is unmodifiable
 	public DiseaseRibbonSummary getDiseaseRibbonSummary(String id,
 														List<String> geneIDs,
-														String includeNegation) {
+														boolean includeNegation) {
```
oblodgett, line 615: These typically should be `Boolean` and not `boolean` and the reason is that what comes in the endpoints are always string and the `Boolean` has a Boolean.parse("String")

oblodgett (approved):

PR #1035, opened 2023-03-22
Fix for changing the type to string vs enum

oblodgett:


cmpich (commented):
Why did you change type from GeneticEntity.CrossReferenceType to String?

oblodgett:
Type has to be a string that is in the resource descriptors file. Which are the displayName not the ENUM name.

agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/PrimaryAnnotatedEntity.java
```diff
@@ -48,7 +47,7 @@ public class PrimaryAnnotatedEntity extends ConditionAnnotation implements Compa
 	@JsonView({View.PrimaryAnnotation.class, View.API.class})
 	protected String url;
 	@JsonView({View.PrimaryAnnotation.class, View.API.class})
-	protected GeneticEntity.CrossReferenceType type;
```
oblodgett, line 51: @cmpich when this type gets serialized needs to match the resource descriptors before this PR would we get the String value of "GENE" not "gene" yes we could lower case it but its not consistent. This way its always the getDisplayName.

cmpich (approved):

PR #1043, opened 2023-04-05
SCRUM-2721: fixing SC

abecerra:
The fix I merged before had the logic inverted!


oblodgett (approved):

agr_api/src/main/java/org/alliancegenome/api/controller/DiseaseController.java
```diff
@@ -289,8 +288,11 @@ public JsonResultResponse<DiseaseAnnotation> getDiseaseAnnotationsByGene(String
 		pagination.addFieldFilter(FieldFilter.ASSOCIATION_TYPE, associationType);
 
 		// TODO: remove when SC data is fixed:
-		if (species.equals("Saccharomyces cerevisiae")) {
-			pagination.addFieldFilter(FieldFilter.SPECIES, "Saccharomyces cerevisiae S288C");
```
oblodgett, line 293: @abecerra this logic is correct... you just need to add the else here.

PR #1048, opened 2023-04-21
SCRUM-2814: upgrade quarkus to 2.16.6

abecerra:


oblodgett:
This ticket was not brought into the sprint nor should we be upgrading these at this time.

PR #1056, opened 2023-05-19
SCRUM-2913 Fix Neo queries

markquintontulloch:
@oblodgett - I've tested locally and this seems to have done the trick.  If I'm missing something then we can discuss on Monday.


oblodgett (approved):
@markquintontulloch yes this is the nature of the change.... although I'm not sure on the TARGET -> TARGETS change but the rest look good, does that cover all pages? We still might find a few more.

oblodgett:
I think [scrum-2727](https://github.com/alliance-genome/agr_java_software/pull/1051/files) when into the wrong branch. @cmpich?

markquintontulloch:
Looks like it went into master, which would explain why I had to switch getCurie() for getReferencedCurie() for xrefs.  Is that why the tests are failing?

oblodgett:
Yes I would believe so... need to talk to christian on why / what we need to do... I would assume could just merge master -> release-6.0.0 (a terrible practice) however I don't think there is much difference at this point.

cmpich:
Correct. SCRUM-2727 should go into release-6.0.0

PR #1060, opened 2023-05-23
SCRUM-2727  filter disease association table.reference column against PubMedID/PubModID

cmpich:


agr_api/src/main/java/org/alliancegenome/api/controller/DiseaseController.java
```diff
@@ -474,12 +475,6 @@ public JsonResultResponse<GeneDiseaseAnnotationDocument> getDiseaseAnnotationsRi
 			}
 		}
 
-		if(filterReference != null) {
-			pagination.addFilterOption("references.crossReferences.curie", QueryParserBase.escape(filterReference) + "&PMID");
```
oblodgett, line 478: The `&` logic should probably be taken out from the backend also.

oblodgett (changes requested):

agr_api/src/main/java/org/alliancegenome/api/controller/DiseaseController.java
cmpich: I am not sure what you mean? The filtering is done on a new field while the old filtering logic has been removed.

oblodgett: `DiseaseESService.java:151` I added that whole else case to handle the `&` and it can be removed.

oblodgett (approved):

PR #1066, opened 2023-06-16
SCRUM-2704 add paralogy chacher infrastructure plus new endpoint for paralogy

cmpich:
…paralogy


oblodgett (changes requested):
Shouldn't the OrthologView be change for ParalogView?

oblodgett (commented):
And the `@JsonView(View.Orthology.class)` be changed for a `@JsonView(View.Paralogy.class)`?

cmpich:
I did not introduce a new view as the existing one, View.Orthology, is just working fine without any changes. If we ever have a need I would split this view up into the two. Maybe, the view should be called View.Homology.class?

cmpich:
Removed ParalogyView class and renamed View.Orthology to View.Homology

oblodgett:
See my slack comments.

oblodgett (approved):

PR #1075, opened 2023-07-03
SCRUM-2823 fix evidence code Column of disease annotation table on gene page

cmpich:
…aryAnnotation object.

### DIFF
--- a/agr_api/src/main/java/org/alliancegenome/api/translators/tdf/DiseaseAnnotationToTdfTranslator.java
+++ b/agr_api/src/main/java/org/alliancegenome/api/translators/tdf/DiseaseAnnotationToTdfTranslator.java
@@ -300,19 +300,32 @@ public class DiseaseAnnotationToTdfTranslator {
 			row.setSource(sourceProvider);
 		}
 */
-		StringJoiner evidenceJoiner = getStringJoiner(annotation, org.alliancegenome.curation_api.model.entities.ontology.ECOTerm::getCurie);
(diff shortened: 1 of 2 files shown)


markquintontulloch (changes requested):
This doesn't make sense to me.  You are getting the URL template from the crossReference dataProvider, but then applying the DOID from the object curie to that template.  If the cross reference on the data provider is not linking to the desired page then that is a data issue.

cmpich:
Removed the commit that changed the crossReference curie setting. That issue will fixed with a data migration script.

agr_api/src/main/java/org/alliancegenome/api/translators/tdf/DiseaseAnnotationToTdfTranslator.java
```diff
@@ -325,6 +325,19 @@ private static StringJoiner getStringJoiner(GeneDiseaseAnnotationDocument annota
 		return evidenceJoiner;
 	}
 
+	private static StringJoiner getStringJoiner(org.alliancegenome.curation_api.model.entities.DiseaseAnnotation annotation, Function<org.alliancegenome.curation_api.model.entities.ontology.ECOTerm, String> function) {
```
oblodgett, line 328: @cmpich do we need the full package path to these classes?

markquintontulloch (approved):

agr_api/src/main/java/org/alliancegenome/api/translators/tdf/DiseaseAnnotationToTdfTranslator.java
cmpich: Unfortunately, yes. We have a DiseaseAnnotation class in the agr_java_software repo.

PR #1081, opened 2023-07-19
SCRUM-2824 adjust mapping of qualifier names

cmpich:


oblodgett (approved):

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
```diff
@@ -139,13 +139,13 @@ private List<GeneDiseaseAnnotationDocument> createGeneDiseaseAnnotationDocuments
...
-	diseaseQualifierMapping.put("sexual_dimorphism", "sexual dimorphism in");
-	diseaseQualifierMapping.put("resistance", "resistance to");
-	diseaseQualifierMapping.put("penetrance", "penetrance of");
+	diseaseQualifierMapping.put("susceptibility_to", "susceptibility to");
+	diseaseQualifierMapping.put("disease_progression_of", "disease progression of");
+	diseaseQualifierMapping.put("severity_of", "severity of");
+	diseaseQualifierMapping.put("onset_of", "onset of");
+	diseaseQualifierMapping.put("sexual_dimorphism_in", "sexual dimorphism in");
+	diseaseQualifierMapping.put("resistance_to", "resistance to");
+	diseaseQualifierMapping.put("penetrance_of", "penetrance of");
```
oblodgett, line 148: Instead of this map could we just switch underscore for space?

cmpich: Works in this case but I am not sure if new values would abide by this. I'll keep it this way for now...

PR #1095, opened 2023-08-29
Fix Cacher to remove unused fields for disease annotations from Neo SCRUM-3228

oblodgett:


markquintontulloch (approved):

cmpich (approved):
Let's see if this change will reduce the size of JSON enough to overcome the error.

PR #1096, opened 2023-08-30
SCRUM-3221 replace old fields on Paralogus object with length, identity, similarity and rank

cmpich:


agr_java_core/src/main/java/org/alliancegenome/neo4j/view/ParalogView.java
```diff
@@ -0,0 +1,87 @@
...
+import lombok.Getter;
+import lombok.Setter;
+import org.alliancegenome.neo4j.entity.node.Gene;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+import java.io.Serializable;
+import java.util.List;
+import java.util.Objects;
+
+@Setter
```
oblodgett: If you use the @Data annotation you dont need to override the equals and compare methods

agr_java_core/src/main/java/org/alliancegenome/neo4j/view/ParalogView.java
```diff
@@ -0,0 +1,87 @@
...
+@Getter
+@Schema(name="OrthologView", description="POJO that represents the Ortholog view")
+@JsonPropertyOrder({"gene", "homologGene", "length", "similarity", "identity", "rank", "predictionMethodsMatched", "predictionMethodsNotMatched", "predictionMethodsNotCalled", "methodCount", "totalMethodCount"})
+public class ParalogView implements Serializable {
```
oblodgett: Can this have a different name than View so that its doesn't get confused with actual views call it summary or something

oblodgett (changes requested):

agr_java_core/src/main/java/org/alliancegenome/neo4j/view/ParalogBean.java
```diff
@@ -0,0 +1,67 @@
...
+	public void setMethodCount(Integer count) {
+	}
+
+	@JsonView(View.Homology.class)
+	@JsonProperty(value = "totalMethodCount")
+	public Integer getTotalMethodCount() {
+		if (predictionMethodsMatched == null && predictionMethodsNotMatched == null)
+			return 0;
+		if (predictionMethodsMatched == null)
+			return predictionMethodsNotMatched.size();
```
oblodgett, line 57: If predictionMethodsNotMatched is null won't this through a NPE? Maybe you want the prior if to be a or vs and?

agr_java_core/src/main/java/org/alliancegenome/neo4j/view/ParalogBean.java
```diff
@@ -0,0 +1,67 @@
...
+import org.alliancegenome.neo4j.entity.node.Gene;
+import org.eclipse.microprofile.openapi.annotations.media.Schema;
+
+import java.io.Serializable;
+import java.util.List;
+
+@Data
+@Schema(name="OrthologView", description="POJO that represents the Paralog view")
+@JsonPropertyOrder({"gene", "homologGene", "length", "similarity", "identity", "rank", "predictionMethodsMatched", "predictionMethodsNotMatched", "predictionMethodsNotCalled", "methodCount", "totalMethodCount"})
+public class ParalogBean implements Serializable {
```
oblodgett, line 16: Bean... implies a "backing bean" from the java yester years. Plus this shouldn't be in the "view" package that is for JsonViews not regular objects.

cmpich: I don't have a better name for it. ParalogSummary is not really describing it either.

agr_java_core/src/main/java/org/alliancegenome/neo4j/view/ParalogBean.java
oblodgett: Is one of these always non null? I realize this is the same code that was in the ortho class but is it correct?

cmpich: I think so but if both can be null this code handles this as well.

oblodgett: The code is using a AND which means that one of them could be null.

markquintontulloch: If they're both null then the code returns 0.  If you get past that then if predictionMethodsMatched is null then predictionMethodsNotMatched cannot be null (otherwise you would have already returned 0).

oblodgett (approved):

PR #1104, opened 2023-09-06
SCRUM-3257 make escaping come after splitting off | selection list

cmpich:


agr_api/src/main/java/org/alliancegenome/api/service/DiseaseESService.java
```diff
@@ -166,6 +152,12 @@ private void generateFilter(BoolQueryBuilder bool, String filterName, String fil
 		//Log.info(bool);
 	}
 
+	private String escapeValue(String value) {
+		value = QueryParser.escape(value);
+		value = value.replaceAll("'", " ");
```
oblodgett, line 156: @cmpich why not pull out the single quotes first then escape everything as the quotes will get escaped and this would leave the backslashes in the text string?

oblodgett (approved):

PR #1105, opened 2023-09-07
strip multiple newlines in synopsis

kkarra:


agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/DataExtractorRepository.java
```diff
@@ -15,7 +15,7 @@ public Result getAllGenes() {
 		query += " OPTIONAL MATCH p2=(g:Gene)-[:ANNOTATED_TO]-(so:SOTerm)";
 		query += " OPTIONAL MATCH p3=(g:Gene)-[:ALSO_KNOWN_AS]-(syn:Synonym)";
 		query += " OPTIONAL MATCH p4=(g:Gene)-[:CROSS_REFERENCE]-(cr:CrossReference)";
-		query += " RETURN g.primaryKey, g.modLocalId, collect(distinct replace(syn.name, ',', '')) as synonyms, collect(distinct cr.name) as crossrefs, g.name, g.symbol,  g.geneSynopsis, g.automatedGeneSynopsis, s.primaryKey, gl.chromosome, gl.start, gl.end, gl.strand, so.name";
+		query += " RETURN g.primaryKey, g.modLocalId, collect(distinct replace(syn.name, ',', '')) as synonyms, collect(distinct cr.name) as crossrefs, g.name, g.symbol,  replace(g.geneSynopsis, '\n', ''), replace(g.automatedGeneSynopsis, '\n', ''), s.primaryKey, gl.chromosome, gl.start, gl.end, gl.strand, so.name";
```
oblodgett, line 18: @kkarra does this work? I would think you need to:
 `replace(g.geneSynopsis, '\n', '') as geneSynopsisStripped` and then pick that up via 
`map.get("geneSynopsisStripped")`

kkarra: @oblodgett  yes it works.

oblodgett: If it works why do you have to strip them twice? Both in the query and in the java?

kkarra: yes, in theory in translator alone should be enough. But, first I tried query then in java, finally a combination that worked, doesn't hurt so left it. We can remove from query I think it will still work but I would have to test again..

oblodgett (approved):

PR #1107, opened 2023-09-13
SCRUM-3280 enable Json on publicationJoins object

cmpich:


agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/DiseaseAnnotation.java
```diff
@@ -76,7 +76,6 @@ public class DiseaseAnnotation extends ConditionAnnotation implements Comparable
 	@JsonView({View.DiseaseAnnotation.class})
 	private List<Gene> orthologyGenes;
 
-	@JsonIgnore
```
oblodgett, line 79: @cmpich yes... I did make this change. There is issues with turning this on.

cmpich: Can we try running this? The approach in not serializing PublicationJoin is not going to work as I need it for the Download logic.

oblodgett (approved):

PR #1111, opened 2023-09-14
Fixed code for indexer

oblodgett:


agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/service/AlleleDiseaseAnnotationService.java
```diff
@@ -59,22 +60,12 @@ public List<AlleleDiseaseAnnotation> getFiltered() {
 		} while(page <= pages);
 		display.finishProcess();
 		alleleRepository.close();
-		
+		geneRepository.close();
 		return ret;
 	}
 
-	private static boolean hasValidAlleles(AlleleDiseaseAnnotation da) {
-		List<String> alleleIds = (new AlleleRepository()).getAllAlleleIDs();
```
oblodgett, line 67: @cmpich Repositories HAVE to be closed otherwise the indexer / cacher hangs.

agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleRepository.java
```diff
@@ -75,11 +102,7 @@ public List<String> getAllAlleleKeys() {
 		return list;
 	}
 
-	private List<String> allAlleleIDs = null;
-
 	public List<String> getAllAlleleIDs() {
-		if (allAlleleIDs != null)
-			return null;
```
oblodgett, line 82: @cmpich the logic here should have been: `if (allAlleleIDs != null) return allAlleleIDs;`

oblodgett: Plus there is no need to cache these as they should only get called a few times.

agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/GeneRepository.java
```diff
@@ -402,23 +357,6 @@ public List<String> getAllAgmKeys() {
...
-		Iterator<Map<String, Object>> i = r.iterator();
-		ArrayList<String> list = new ArrayList<>();
-
-		while (i.hasNext()) {
-			Map<String, Object> map2 = i.next();
-			list.add((String) map2.get("g.primaryKey"));
-		}
-		return list;
-	}
-
```
oblodgett, line 421: removed unused code.

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/service/AlleleDiseaseAnnotationService.java
oblodgett: This is due to a persistent connection to Neo.

agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/AlleleRepository.java
cmpich: But I don't think this causes too much memory strain on the indexer. This will increase the timing of the indexer.

cmpich (approved):

PR #1120, opened 2023-09-26
SCRUM-3305 fix variant category filtering bug

cmpich:


agr_java_core/src/main/java/org/alliancegenome/core/variant/service/AlleleVariantIndexService.java
```diff
@@ -304,8 +294,7 @@ public BoolQueryBuilder buildBoolQuery(String id, Pagination pagination) {
 					}
 
 					if (e.getKey().toString().equalsIgnoreCase("allele_category")) {
-						if (e.getValue().toString().split("\\|").length < 3)
-							queryBuilder.filter(QueryBuilders.termsQuery("alterationType.keyword", e.getValue().toString().split("\\|")));
+						queryBuilder.filter(QueryBuilders.termsQuery("alterationType.keyword", e.getValue().toString().split("\\|")));
```
oblodgett, line 297: Was alteration type a two digit code before? Seems weird if it was.

oblodgett:
We have already built test, so this will need to wait. But good news we will be able to deploy quicker with this new pipeline.

oblodgett (approved):

agr_java_core/src/main/java/org/alliancegenome/core/variant/service/AlleleVariantIndexService.java
cmpich: I cannot remember but it looks like the list of values used to be less than 3.

PR #1121, opened 2023-10-03
SCRUM-3286 Update curation library to v0.25.0

markquintontulloch:
@oblodgett @cmpich 
v0.25.0 brings in a breaking change for disease annotations in that `diseaseRelation` has been renamed to `relation`.  If my understanding of the main site is correct (probably not) then if we keep `diseaseRelation` in the DiseaseAnnotationDocument then the UI should be unaffected by this change and the only changes required are the changes made here to the DiseaseAnnotationCurationIndexer.

However, I don't know if we also want to change the document to bring it in line with the latest LinkML version and make any required UI changes.  Given there's a bunch of stuff going on in the document that's not covered by the LinkML model then I'm guessing this isn't really necessary.


markquintontulloch:
I take it back!  That change isn't coming until a later version of the library.

oblodgett (approved):

PR #1122, opened 2023-10-08
Initial Code for quarkus 3.4 upgrade

oblodgett:


agr_api/src/main/java/org/alliancegenome/api/controller/OrthologyController.java
```diff
@@ -18,27 +15,25 @@
...
 
 	@Inject OrthologyCacheRepository service;
 
 	@Inject GeneController controller;
-	
+
 	private static OrthologousRepository orthoRepo = new OrthologousRepository();
 
 	@Override
-	public JsonResultResponse<HomologView> getDoubleSpeciesOrthology(String taxonIDOne,
```
cmpich, line 36: Why are you changing this? I like to have the various arguments underneath each other for ease of readability.

oblodgett: This is auto formatting in eclipse.

markquintontulloch (approved):

agr_api/src/main/java/org/alliancegenome/api/controller/OrthologyController.java
cmpich: But it is not a good formatting. Can you change that 
default behavior in eclipse?

PR #1130, opened 2023-10-23
Bump curation version to v0.26.0 (SCRUM-3367)

mluypaert:


markquintontulloch:
Do we really want to change diseaseRelation to relation everywhere?  This will presumably require UI changes too.  Or do we just want to change it in the indexer where pulling the information from the curation system and creating the search documents?

mluypaert:
> Do we really want to change diseaseRelation to relation everywhere? This will presumably require UI changes too. Or do we just want to change it in the indexer where pulling the information from the curation system and creating the search documents?

Well why did we change the property name but not the display name in the curation application? Changing backend name but not display name is an anti-pattern to my opinion and should be avoided unless there is a good reason to.

markquintontulloch:
> > Do we really want to change diseaseRelation to relation everywhere? This will presumably require UI changes too. Or do we just want to change it in the indexer where pulling the information from the curation system and creating the search documents?
> 
> Well why did we change the property name but not the display name in the curation application? Changing backend name but not display name is an anti-pattern to my opinion and should be avoided unless there is a good reason to.

We changed the property name in the curation system because it was at odds with the LinkML model.  The search documents do not follow the LinkML model.  I agree that consistency is good, but think that this requires a bit more thought as to what we change and when.  For example, why replace `diseaseRelationNegation` with `relationNegation` when this is something that is constructed for the search document, has nothing to do with the LinkmL model, and will probably result in more UI failures if changed?

markquintontulloch:
`diseaseRelationNegation` is constructed from the `negated` field and the `relation` field.  It is not part of the LinkML model and, since it is specific to disease annotations, I don't see any compelling reason to change the name.  From a quick check, the UI code only seems to refer to `diseaseRelationNegation` and never directly to `diseaseRelation`/`relation`.  As such, I'd be inclined to keep `diseaseRelationNegation` as it is so that this doesn't cause breaking changes in the UI repo.

That said, I'm not particularly familiar with the code in either this repo or the UI one so I will leave it to @cmpich or @oblodgett to approve this PR or request changes.

agr_java_core/src/main/java/org/alliancegenome/api/entity/DiseaseAnnotationDocument.java
```diff
@@ -22,13 +22,13 @@
...
 	private String uniqueId;
-	private VocabularyTerm diseaseRelation;
+	private VocabularyTerm relation;
 	private DOTerm object;
 	private List<ECOTerm> evidenceCodes;
 	private Reference singleReference;
 	private Set<Reference> references;
 
-	private String diseaseRelationNegation;
+	private String relationNegation;
```
oblodgett: I realize that the way this was initially implemented was with certain names. I would prefer not adding "negation" to everything. Just call it what it is: "relation". And if we need a "generated" string then call it that "relationText" or "relationGeneratedText" etc. And if we need to "negate" the text then call a method that has an appropriate name "negateRelationText".
We use the wrapper class "DiseaseAnnotationDocument" for convenience. That way the documents going into ES are much smaller than the object coming out of the curation system. We use only the fields needed to show the table and nothing more. To me this change is fine the way it is, however I would like to see better naming but I will leave that up to @markquintontulloch

markquintontulloch: Yeah, I'm not a fan of diseaseRelationNegation / relationNegation either - I think this implies it is always the negated version whereas it's actually the generated string for both positive and negative relations.
If we're going to cause breaking changes in the UI then I'd prefer renaming to something more descriptive as @oblodgett describes above.  However, I'd prefer not to rename it to "relationNegation" as currently proposed since it is no more descriptive, has nothing to do with the LinkML, and will require UI changes.

oblodgett: The UI changes for this shouldn't be that extensive, as can be seen here: https://github.com/search?q=repo%3Aalliance-genome%2Fagr_ui%20diseaseRelation&type=code and I would also agree that it should reflect LinkML whenever possible.

markquintontulloch: I hadn't spotted the direct reference to `diseaseRelation` in the UI code in AnnotatedEntitiesPopupCuration.js.  Given that, my vote would definitely be to replace `diseaseRelationNegation` with something more descriptive than `relationNegation` and then replace `diseaseRelation` with `relation` everywhere else as in the current PR.

agr_java_core/src/main/java/org/alliancegenome/api/entity/DiseaseAnnotationDocument.java
```diff
@@ -28,7 +28,7 @@ public class DiseaseAnnotationDocument extends SearchableItemDocument {
 	private Reference singleReference;
 	private Set<Reference> references;
 
-	private String relationNegation;
+	private String negationInclRelation;
```
mluypaert: @oblodgett @markquintontulloch Now updated, and I've immediately opened a PR in the UI repo as well to match these changes: alliance-genome/agr_ui#1213

agr_java_core/src/main/java/org/alliancegenome/api/entity/DiseaseAnnotationDocument.java
```diff
@@ -22,13 +22,13 @@
...
 	private String uniqueId;
-	private VocabularyTerm diseaseRelation;
+	private VocabularyTerm relation;
 	private DOTerm object;
 	private List<ECOTerm> evidenceCodes;
 	private Reference singleReference;
 	private Set<Reference> references;
 
-	private String diseaseRelationNegation;
+	private String negationInclRelation;
```
oblodgett: @markquintontulloch what does "Incl" mean? Include? if so just spell it out? Variable names are cheap confusion is expensive.

mluypaert: @oblodgett negationInclusiveRelation, but I found that to be a bit too long as a variable name. Incl seems common enough as an abbreviation to me to be acceptable here.

markquintontulloch: Ask @mluypaert, but I assume so!  Can't say I'm a huge fan of the new field name - I'd have preferred relationIncludingNegation, or just relationText, as it's a string describing the relation.

oblodgett: Ya I am in favor of the relationText also.

mluypaert: @markquintontulloch The problem with using "Text" as a description is that is doens't explain anything about the content, nor does this string represent a display text, this string only has an additional `_not_` inserted into the relation string if it is negated, while the display text would essentially be replacing all _ with spaces.

oblodgett: is the field ALWAYS the negated version of the relation? Maybe thats the confusion here? If so just call it that. (negatedRelationText)

mluypaert: @oblodgett no, only when negated is true, see the code that generates it:
https://github.com/alliance-genome/agr_java_software/pull/1130/files#diff-8272ea294f7bd959d0f02f20fe9d495da2a40ad9e9a8669bbfaa71cfbb5402f2L164-R167

oblodgett: @mluypaert right thats what I suspected. So I would just call it that "relationText".

markquintontulloch: Not sure what you mean by it not representing a display text.  Isn't that exactly what it is - the text that is displayed for a relation, which takes the negated field into account?

oblodgett: ya... I would be fine with displayRelationText also... which tells me its generated.

mluypaert: @markquintontulloch @oblodgett  No, because the text being displayed to users (in the website) is `is implicated in` or `is NOT implicated in`, while the variable here will hold `is_implicated_in` or `is_not_implicated_in`. Hence it's closer to the relation value than it is to the display text.

mluypaert: This is not text, merely the relation value including the negation property.

mluypaert: @oblodgett @markquintontulloch Does `relationNegationMerge` sound any better (than `negationInclRelation` or `negationInclusiveRelation`)?

oblodgett: It doesn't need to include "negation". From my point if we generate something from obsolete or negated or whatever criteria that doesn't need to be included in the name.... just a way that the name indicates that its a generated field.

markquintontulloch: I'd go with something like `relationString` or `generatedRelationString` if you're not happy with `relationText`.

oblodgett: So I do agree that "relationText" would be misleading if we are going to modify it to include other things. So to me "generatedRelationText" would be best then the variable name is independent of its content.

agr_java_core/src/main/java/org/alliancegenome/api/entity/DiseaseAnnotationDocument.java
```diff
@@ -22,13 +22,13 @@
...
 	private String uniqueId;
-	private VocabularyTerm diseaseRelation;
+	private VocabularyTerm relation;
 	private DOTerm object;
 	private List<ECOTerm> evidenceCodes;
 	private Reference singleReference;
 	private Set<Reference> references;
 
-	private String diseaseRelationNegation;
+	private String generatedRelationString;
```
mluypaert, line 31: @oblodgett @markquintontulloch Better like this?

oblodgett: Yes thanks... this makes a lot more sense to me.

oblodgett (approved):

mluypaert:
Alright, thanks, mind approving the UI PR as well @oblodgett? (alliance-genome/agr_ui#1213)

PR #1134, opened 2023-11-14
SCRUM-2393 make inferred or asserted AlleleDAs have is_implicated_in …

cmpich:
…relationships

### DIFF
--- a/agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
+++ b/agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
@@ -75,15 +75,13 @@ public class DiseaseAnnotationCurationIndexer extends Indexer {
 		diseaseRepository = new DiseaseRepository();
 		closureMap = diseaseRepository.getDOClosureChildMapping();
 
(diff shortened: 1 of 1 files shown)


agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
```diff
@@ -75,13 +75,15 @@ protected void index() {
 		diseaseRepository = new DiseaseRepository();
 		closureMap = diseaseRepository.getDOClosureChildMapping();
 
-		indexGenes();
+		//indexGenes();
```
oblodgett: @cmpich do you really want to turn off indexing genes?

oblodgett (changes requested):
Do we need to turn off Gene indexing?

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
cmpich: Good catch. Undid it.

oblodgett (approved):

PR #1147, opened 2024-01-05
SCRUM-3315: create a new list of evidence codes

abecerra:


oblodgett (approved):

oblodgett (changes requested):
This is still not right.

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
```diff
@@ -133,7 +134,11 @@ private List<GeneDiseaseAnnotationDocument> createGeneDiseaseAnnotationDocuments
 					gdad.setParentSlimIDs(closureMap.get(da.getObject().getCurie()));
 					lookup.put(key, gdad);
 				}
-				gdad.setEvidenceCodes(da.getEvidenceCodes());
+
```
oblodgett, line 137: So the current list that is in the `GeneDiseaseAnnotationDocument gdad` needs to be added also, not overridden. So you need to take the list that is in gdad and add those to the map first then add the ones from da and then do the setEvidenceCodes.

abecerra: ok

oblodgett (approved):

PR #1152, opened 2024-01-31
SCRUM-3361: add secondary data provider

abecerra:


cmpich (approved):
Looks good.

oblodgett (approved):

cmpich (changes requested):
This is an AND operation on main data provider and secondary data provider with the same value. That is not going to work.

Need to make an OR operation between the two fields.

abecerra:
Do we need a should query?

https://jolicode.com/blog/how-to-build-a-or-condition-in-elasticsearch-query-dsl

PR #1156, opened 2024-02-16
SCRUM-3697

cmpich:


agr_java_core/src/main/java/org/alliancegenome/neo4j/repository/GeneRepository.java
```diff
@@ -76,19 +57,42 @@ public Gene getOneGene(String primaryKey) {
...
+			       COLLECT(crossRef) AS CrossReferences
+			""";
 
 		Iterable<Gene> genes = query(query, map);
-		for (Gene g : genes) {
-			if (g.getPrimaryKey().equals(primaryKey)) {
-				return g;
-			}
+		if (genes.iterator().hasNext()) {
+			Gene gene = genes.iterator().next();
```
oblodgett, line 82: @cmpich you need to do the `if (gene.getPrimaryKey().equals(primaryKey)) {` check sometimes for whatever reason two genes match and this iterator will be looking at the wrong one.

oblodgett (changes requested):

oblodgett (approved):

PR #1160, opened 2024-03-04
Sort

cmpich:


agr_java_core/src/main/java/org/alliancegenome/api/entity/GeneDiseaseAnnotationDocument.java
```diff
@@ -18,4 +20,26 @@ public class GeneDiseaseAnnotationDocument extends DiseaseAnnotationDocument {
...
 		setCategory("gene_disease_annotation");
 	}
+
+	// 1 true
+	// 0 false
+	@JsonIgnore
+	private boolean isViaOrthologyAnnotation;
+
+	@JsonView({View.DiseaseAnnotationAll.class})
+	@JsonProperty("viaOrthologyOrder")
```
oblodgett: These jsonProperties are the same name as the method so they should be removed.

agr_java_core/src/main/java/org/alliancegenome/cache/repository/helper/JsonResultResponse.java
```diff
@@ -101,6 +102,7 @@ public void addDistinctFieldValueSupplementalData(Map object) {
 		supplementalData.put(DISTINCT_FIELD_VALUES, object);
 	}
 
+	@JsonIgnore
```
oblodgett: Isn't this used all around the site? Why would we need a @JsonIgnore?

cmpich: I wonder if any UI component is using this. The distinctFieldValues should only be used as part of the supplementalData, otherwise it is duplicated in the returned payload.

cmpich: The UI solely accesses this object from the supplementalData object and makes this top-level output superfluous.

oblodgett: The UI is using this field:
```
components/dataTable/DataTable.js:      const distinctFieldValues = Array.isArray(column.filterable) ?
components/dataTable/DataTable.js:      if (distinctFieldValues && distinctFieldValues.length > 0) {
components/dataTable/DataTable.js:            options={distinctFieldValues}
components/dataTable/DataTable.js:        //if filter is a checkbox dropdown, but there is are no distinctFieldValues then show DropdownNoDataFilter instead
components/dataTable/utils.js:  const {distinctFieldValues = {}} = response.supplementalData || {};
components/dataTable/utils.js:  return (distinctFieldValues[field] || [])
```

oblodgett: Only one case where its pulling it from the supplementalData? Maybe I am missing something?

agr_java_core/src/main/java/org/alliancegenome/api/entity/GeneDiseaseAnnotationDocument.java
```diff
@@ -18,4 +20,24 @@ public class GeneDiseaseAnnotationDocument extends DiseaseAnnotationDocument {
...
+	public int getViaOrthologyOrder() {
+		return isViaOrthologyAnnotation ? 1 : 0;
+	}
+
+	@JsonView({View.DiseaseAnnotationAll.class})
+	public void setViaOrthologyOrder(int order) {
+		isViaOrthologyAnnotation = order == 1;
+	}
+
+	@Override
```
oblodgett: Can we remove the toString this should be done via `@ToString` at the class level.

oblodgett (changes requested):

agr_java_core/src/main/java/org/alliancegenome/api/entity/GeneDiseaseAnnotationDocument.java
cmpich: Removed.

agr_java_core/src/main/java/org/alliancegenome/cache/repository/helper/JsonResultResponse.java
cmpich: Adam pointed out that there is a util function that retrieves the distinctFieldValues:
```

export const getDistinctFieldValue = (response, field) => {
  response = response || {};
  const {distinctFieldValues = {}} = response.supplementalData || {};
  return (distinctFieldValues[field] || [])
    .sort(compareAlphabeticalCaseInsensitive)
     // TODO: remove when backend is fixed, see https://agr-jira.atlassian.net/browse/SCRUM-2649
    .map(simplifySpeciesNameSC)
    .filter((value) => (
      value && value.trim()
    ));
};
```

cmpich: And if there is a reference in the UI and we should change it.

oblodgett: Ok... this method is only used in the Integration tests in the Java software. I propose that we change its name from `get` to `retrieve` then we don't need the `@JsonIgnore`

oblodgett (approved):

PR #1161, opened 2024-03-06
Add IEA to via orthology DAs

cmpich:


agr_indexer/src/main/java/org/alliancegenome/indexer/rest/interfaces/EcoTermRESTInterface.java
```diff
@@ -0,0 +1,19 @@
...
+
+@Path("/ecoterm")
+@Produces({"application/json"})
+@Consumes({"application/json"})
+public interface EcoTermRESTInterface {
+
+	@GET
+	@Path("/{curie}")
+	@JsonView({View.FieldsAndLists.class})
+	ObjectResponse<ECOTerm> find(@PathParam("curie") String curie);
```
oblodgett, line 17: Does the interface in the curation api not work for this?

cmpich: I have not tried but we always had trouble with it.

oblodgett: Due to Json view issues but that shouldn't be a problem here.

oblodgett: We wanted a different view so we had to reimplement the interfaces. However in the end we made the change on the Curation side. So I would guess we don't need the Annotation Serivces / Interfaces anymore either.

oblodgett (approved):

PR #1167, opened 2024-03-13
SCRUM-3649 replace newlines with white spaces, show main entity genet…

cmpich:
…ic profile in download file for via-orthology records

### DIFF
--- a/agr_api/src/main/java/org/alliancegenome/api/translators/tdf/DiseaseAnnotationToTdfTranslator.java
+++ b/agr_api/src/main/java/org/alliancegenome/api/translators/tdf/DiseaseAnnotationToTdfTranslator.java
@@ -3,7 +3,6 @@ package org.alliancegenome.api.translators.tdf;
 import org.alliancegenome.api.entity.AlleleDiseaseAnnotationDocument;
 import org.alliancegenome.api.entity.DiseaseAnnotationDocument;
 import org.alliancegenome.api.entity.GeneDiseaseAnnotationDocument;
-import org.alliancegenome.core.config.ConfigHelper;
 import org.alliancegenome.core.translators.tdf.DiseaseDownloadRow;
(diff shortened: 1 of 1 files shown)


agr_api/src/main/java/org/alliancegenome/api/translators/tdf/DiseaseAnnotationToTdfTranslator.java
```diff
@@ -229,9 +233,9 @@ private static void extracted(DiseaseAnnotationDocument annotation, org.alliance
 		DataProvider dataProvider = primaryAnnotation.getDataProvider();
 		if (dataProvider != null && dataProvider.getCrossReference() != null) {
 			String urlTemplate = dataProvider.getCrossReference().getResourceDescriptorPage().getUrlTemplate();
-			if(urlExpceptionHandler.contains(dataProvider.getSourceOrganization().getAbbreviation())){
+			if (urlExpceptionHandler.contains(dataProvider.getSourceOrganization().getAbbreviation())) {
 				// remove the prefix in the template as the prefix is already in the curie.
-				urlTemplate = urlTemplate.replace(dataProvider.getSourceOrganization().getAbbreviation()+":","");
+				urlTemplate = urlTemplate.replace(dataProvider.getSourceOrganization().getAbbreviation() + ":", "");
 			}
```
oblodgett, line 239: this code needs to be in the url template object basically pass the id in and get a url back you will need this in multiple places

cmpich: I agree, but that has to happpen in the curation repo first until it can get used here.
I'd like to get his out before sprint review.

oblodgett (approved):
Can we put in a PR in the curation repo for this?

PR #1196, opened 2024-05-28
SCRUM-4092 make all disease qualifier names be their original names i…

cmpich:
…ncluding underscores


abecerra (approved):

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
```diff
@@ -249,11 +249,13 @@ private DiseaseAnnotation createImplicatedDA(DiseaseAnnotation da) {
 			AlleleDiseaseAnnotation alleleDA = new AlleleDiseaseAnnotation();
 			alleleDA.setDiseaseAnnotationSubject(alleleAnno.getDiseaseAnnotationSubject());
 			implicatedDA = alleleDA;
+		} else {
+			throw new RuntimeException("CreateImplicatedDA() Disease Annotations can only be used for AGM DAs or Allele DAs.");
```
oblodgett: This runtime exception can be brought up to line 203.

oblodgett (changes requested):
The if else should be brought up to the earlier if/else so that the instance checking all happens in one spot.

oblodgett (approved):

PR #1201, opened 2024-05-30
SCRUM-4099 fixup species name in filtering option

cmpich:


agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/SpeciesType.java
```diff
@@ -170,6 +170,10 @@ public static String getTaxonId(String species) {
 		return species;
 	}
 
+	// Fix up species name with yeast correction
+	public static String getSpeciesNameCorrected(String name){
```
oblodgett, line 174: Given that this is yeast only, I would suggest a different name for this method.

agr_api/src/main/java/org/alliancegenome/api/controller/DiseaseController.java
```diff
@@ -317,6 +317,18 @@ public JsonResultResponse<GeneDiseaseAnnotationDocument> getDiseaseAnnotationsBy
 		}
 	}
 
+	/**
+	 * replace Saccharomyces cerevisiae with Saccharomyces cerevisiae S288C
```
oblodgett: This method just splits the input into a list? So this comment doesn't fit what the method does?

agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/SpeciesType.java
cmpich: I want to keep this method generic, in case other species do something similar.

oblodgett (approved):

PR #1203, opened 2024-06-07
Updated variants to use versioned files vs hardcoded ones

oblodgett:


agr_cacher/src/main/java/org/alliancegenome/cacher/cachers/AlleleCacher.java
```diff
@@ -209,39 +211,43 @@ private void readAllFileMetaData() {
...
-			executor.submit(creator);
-			//Thread.sleep(1000);
-
-			executor.shutdown();
-			while (!executor.isTerminated()) {
-				Thread.sleep(1000);
-			}
+			
+//			ExecutorService executor = Executors.newFixedThreadPool(VariantConfigHelper.getSourceDocumentCreatorThreads());
+//
```
cmpich: Why is this commented out? Can that be removed?

oblodgett: I just removed it all.

markquintontulloch (approved):

PR #1204, opened 2024-06-11
SCRUM-4150 validate various entities of disease annotation for being …

cmpich:
…obsolete


markquintontulloch (commented):
Do we need to add 'with' genes to the list of stuff to be checked?

oblodgett (approved):

markquintontulloch:
Does anything extra need to be done for the inferred annotations?

oblodgett:
Sorry approved this too fast, isn't there an asserted Gene also?

markquintontulloch (changes requested):
Pretty sure 'with' needs to be added to the list of things to validate, otherwise looks good.  Don't know if any extra logic is required for the inferred annotations.

cmpich:
The inferred/asserted genes and alleles are all included already in the logic.
The inferred annotations are using entities that have been checked before they get created.
I added the With-Genes.

cmpich:
Do not merge this until 7.2.0 is deployed and 7.3.0 is prepared on stage repo.

markquintontulloch (approved):

markquintontulloch:
Just thought, shouldn't we also be checking internal status on all of these things too?

PR #1227, opened 2025-01-31
SCRUM-4823 Update hard-coded assembly names to current versions

markquintontulloch:


oblodgett (approved):

oblodgett:
this might be effected by the names that were put into neo we might want to check those first before bringing this PR in. Also is this PR for 8.0.0 or 8.1.0?

markquintontulloch:
@oblodgett Not sure what you mean by "names that were put into Neo" - where do I need to check for that?  As for 8.0.0 or 8.1.0 - that's probably a question for @chris-grove.

oblodgett:
In Neo there is going to be a node I think Assembly that will have a "primaryKey/Id" with this name. They will need to match and that node is created via the loader.... it may be that the node name has already changed, but just need to make sure. I know there is comparisions in the code that look at the Assembly name and compare it to that Java Enum name.
... here is what is in Neo

```
{"dataProvider":"Alliance of Genome Resources","primaryKey":"ASM985889v3"}
{"dataProvider":"SGD","primaryKey":"R64-5-1"}
{"dataProvider":"Xenbase","primaryKey":"XT9.1"}
{"dataProvider":"Xenbase","primaryKey":"XL9.2"}
{"dataProvider":"FB","primaryKey":"R6"}
{"dataProvider":"ZFIN","primaryKey":"GRCz11"}
{"dataProvider":"RGD","primaryKey":"GRCh38"}
{"dataProvider":"RGD","primaryKey":"mRatBN7.2"}
{"dataProvider":"WB","primaryKey":"WBcel235"}
{"dataProvider":"MGI","primaryKey":"GRCm39"}
{"dataProvider":"MGI","primaryKey":""} 
```
So I would guess everything is ok... ?

PR #1232, opened 2025-02-19
SCRUM-4785 & SCRUM-4786 Index interactions from curation system

markquintontulloch:
@oblodgett Have just copied the cryptic numbers in the IndexerConfig from the DiseaseAnnotationMlIndexer - presumably these need to change.

### DIFF
--- a/agr_indexer/src/main/java/org/alliancegenome/indexer/config/IndexerConfig.java
+++ b/agr_indexer/src/main/java/org/alliancegenome/indexer/config/IndexerConfig.java
@@ -7,6 +7,8 @@ import org.alliancegenome.indexer.indexers.GeneIndexer;
 import org.alliancegenome.indexer.indexers.GoIndexer;
 import org.alliancegenome.indexer.indexers.ModelIndexer;
 import org.alliancegenome.indexer.indexers.curation.DiseaseAnnotationCurationIndexer;
+import org.alliancegenome.indexer.indexers.curation.GeneGeneticInteractionCurationIndexer;
+import org.alliancegenome.indexer.indexers.curation.GeneMolecularInteractionCurationIndexer;
 
 public enum IndexerConfig {
 
@@ -23,6 +25,8 @@ public enum IndexerConfig {
 	GoIndexer("go", GoIndexer.class, 4, 914, 914, 8, 1),				// (44087,	262, 460247, 1093)
 	ModelIndexer("model", ModelIndexer.class, 4, 1426, 1426, 8, 1),		   // (132447, 314,	 18593,	 701)
 	DiseaseAnnotationMlIndexer("diseaseAnnotation", DiseaseAnnotationCurationIndexer.class, 1, 1426, 1426, 2, 1),
+	GeneGeneticInteractionIndexers("geneGeneticInteraction", GeneGeneticInteractionCurationIndexer.class, 1, 1426, 1426, 2, 1),
+	GeneMolecularInteractionIndexers("geneMolecularInteraction", GeneMolecularInteractionCurationIndexer.class, 1, 1426, 1426, 2, 1),
 	// still not implemented  VariantIndexer("variant", VariantIndexer.class, 2, 3000, 400, 4, 4)
 	;
 
--- a/agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/Indexer.java
+++ b/agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/Indexer.java
@@ -50,7 +50,7 @@ public abstract class Indexer extends Thread {
 	protected DecimalFormat df = new DecimalFormat("#");
 	protected ObjectMapper om = new ObjectMapper();
 
-	protected ProcessDisplayHelper display = new ProcessDisplayHelper();
+	private ProcessDisplayHelper display = new ProcessDisplayHelper();
 	private StatsCollector stats = new StatsCollector();
 
 	protected Map<String, Double> popularityScore;
--- a/agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/Indexer.java
+++ b/agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/Indexer.java
@@ -52,7 +52,7 @@ public abstract class Indexer extends Thread {
 	protected DecimalFormat df = new DecimalFormat("#");
 	protected ObjectMapper om = new ObjectMapper();
 
-	private ProcessDisplayHelper display = new ProcessDisplayHelper();
+	protected ProcessDisplayHelper display = new ProcessDisplayHelper();
 	private StatsCollector stats = new StatsCollector();
 
 	protected Map<String, Double> popularityScore;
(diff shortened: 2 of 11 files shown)


agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/service/helpers/GeneInteractionHelper.java
```diff
@@ -0,0 +1,49 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class GeneInteractionHelper {
+
+	
+	private ObjectMapper objectMapper = new ObjectMapper().configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false);
+	
+	public GeneGeneticInteraction generateReverseInteraction(GeneGeneticInteraction forwardInteraction) throws IOException {
+		
+		GeneGeneticInteraction newInteraction = objectMapper.readValue(objectMapper.writeValueAsString(forwardInteraction), GeneGeneticInteraction.class);
```
oblodgett: This is an interesting idea for "cloning" an interaction. I just wonder if it might cause problems in the future with timestamps etc.

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/Indexer.java
```diff
@@ -52,7 +52,7 @@ public abstract class Indexer extends Thread {
 	protected DecimalFormat df = new DecimalFormat("#");
 	protected ObjectMapper om = new ObjectMapper();
 
-	private ProcessDisplayHelper display = new ProcessDisplayHelper();
+	protected ProcessDisplayHelper display = new ProcessDisplayHelper();
```
oblodgett: Does this need to be changed? I don't see it being used in this code?

agr_indexer/src/main/java/org/alliancegenome/indexer/config/IndexerConfig.java
```diff
@@ -23,6 +25,8 @@ public enum IndexerConfig {
 	GoIndexer("go", GoIndexer.class, 4, 914, 914, 8, 1),				// (44087,	262, 460247, 1093)
 	ModelIndexer("model", ModelIndexer.class, 4, 1426, 1426, 8, 1),		   // (132447, 314,	 18593,	 701)
 	DiseaseAnnotationMlIndexer("diseaseAnnotation", DiseaseAnnotationCurationIndexer.class, 1, 1426, 1426, 2, 1),
+	GeneGeneticInteractionIndexers("geneGeneticInteraction", GeneGeneticInteractionCurationIndexer.class, 1, 1426, 1426, 2, 1),
+	GeneMolecularInteractionIndexers("geneMolecularInteraction", GeneMolecularInteractionCurationIndexer.class, 1, 1426, 1426, 2, 1),
```
oblodgett: You might want to change the thread count here to 4 and set the buffer to a more whole number.

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/service/helpers/GeneInteractionHelper.java
markquintontulloch: Yeah, it'll break if any OffsetDateTime fields come through - but they don't at present with the ForPublic view.  Not sure how likely we are to bring timestamps through on the interactions in the future.

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/Indexer.java
markquintontulloch: Reverted

agr_indexer/src/main/java/org/alliancegenome/indexer/config/IndexerConfig.java
markquintontulloch: Updated

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/service/helpers/GeneInteractionHelper.java
markquintontulloch: I take that back - I see we have the ForPublic view on `dateCreated`.  Will rewrite this code

markquintontulloch: @oblodgett Now changed

oblodgett: Where I was going with this is you might need to have the following: 
```java
mapper.registerModule(new JavaTimeModule());
```
in order for the dates to "copy" correctly

however what would probably be better is to use the mapper that is defined for the indexer's interaction with the curation API:
```java
mapper = RestConfig.config.getJacksonObjectMapperFactory().createObjectMapper();
```

markquintontulloch: @oblodgett - updated now

markquintontulloch: Actually, still not working.  Getting the following error:

`com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Java 8 date/time type 'java.time.OffsetDateTime' not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310" to enable handling (through reference chain: org.alliancegenome.api.entity.GeneGeneticInteractionDocument["geneGeneticInteraction"]->org.alliancegenome.curation_api.model.entities.GeneGeneticInteraction["geneAssociationSubject"]->org.alliancegenome.curation_api.model.entities.Gene["dateCreated"])
`

which is weird because the `createObjectMapper()` method has `mapper.registerModule(new JavaTimeModule();`

oblodgett (approved):

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/Indexer.java
```diff
@@ -212,6 +213,7 @@ public void run() {
 	protected abstract void startSingleThread(LinkedBlockingDeque<String> queue);
 
 	protected ObjectMapper customizeObjectMapper(ObjectMapper objectMapper) {
+		objectMapper.registerModule(new JavaTimeModule());
```
oblodgett, line 216: @markquintontulloch I didn't catch this... this effects ALL indexers. I think what needs to happen like in the disease indexer there is an override:
```java
	@Override
	protected ObjectMapper customizeObjectMapper(ObjectMapper objectMapper) {
		return RestConfig.config.getJacksonObjectMapperFactory().createObjectMapper();
	}
```

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/service/helpers/GeneInteractionHelper.java
```diff
@@ -0,0 +1,46 @@
...
+		reverseInteraction.setGeneGeneAssociationObject(forwardInteraction.getGeneAssociationSubject());
+		
+		reverseInteraction.setInteractorARole(forwardInteraction.getInteractorBRole());
+		reverseInteraction.setInteractorBRole(forwardInteraction.getInteractorARole());
+		
+		reverseInteraction.setInteractorAType(forwardInteraction.getInteractorBType());
+		reverseInteraction.setInteractorBType(forwardInteraction.getInteractorAType());
+		
+		return reverseInteraction;
+	}
```
oblodgett, line 45: PS all this code could just go in the service as the service's in the indexer are not  really attached to anything

PR #1234, opened 2025-02-21
Removed unused mapper

oblodgett:
@cmpich why was this code added?

https://github.com/alliance-genome/agr_java_software/commit/6b26b9aa1005614fe7f688dd1c8d07fe9ca1c915


cmpich (approved):
I don't know how and why it got here.

PR #1270, opened 2025-03-12
SCRUM-4795, SCRUM-4825 - Use sort indexes

markquintontulloch:


oblodgett (approved):
Do these not get used later in the code?

PR #1296, opened 2025-03-28
SCRUM-4795 & SCRUM-4925 Fix interaction table source filters

markquintontulloch:


oblodgett (commented):
So... given these long chains of fields, would it be better to just bring those value up into Fields that hang directly off the document so your filter would just be `source` and it could be a Single String or Set of Strings this way making changes only has to happen in the builder and then the field just gets new values?

markquintontulloch:
> So... given these long chains of fields, would it be better to just bring those value up into Fields that hang directly off the document so your filter would just be `source` and it could be a Single String or Set of Strings this way making changes only has to happen in the builder and then the field just gets new values?

@oblodgett Not sure I follow.  Do you mean create new fields on the documents that contain all values that each field can be filtered on, and use those for filtering?  If so, I'm not sure I like that - I'd rather be able to adjust filtering in the API without having to reindex everything.

agr_api/src/main/java/org/alliancegenome/api/controller/GeneController.java
```diff
@@ -442,7 +442,7 @@ public Response getMolecularInteractionsDownload(String id, String sortBy, Strin
 		Pagination pagination = new Pagination(1, 150000, sortBy, asc);
 		pagination.addFilterOption("geneMolecularInteraction.interactorAType.name.keyword", moleculeType);
 		pagination.addFilterOption("geneMolecularInteraction.geneGeneAssociationObject.geneSymbol.displayText", interactorGeneSymbol);
-		pagination.addFilterOption("geneMolecularInteraction.interactionIdORgeneMolecularInteraction.aggregationDatabase.nameORgeneMolecularInteraction.interactionSource.name", source);
+		pagination.addFilterOption("geneMolecularInteraction.interactionIdORgeneMolecularInteraction.aggregationDatabase.nameORgeneMolecularInteraction.interactionSource.nameORgeneMolecularInteraction.crossReferences.displayName", source);
```
oblodgett, line 445: @markquintontulloch I am guessing that `interactionIdORgeneMolecularInteraction` is a method call and not a field, so it should be @JsonIgnored and not in the document to start with. Where are these fields `interactionIdORgeneMolecularInteraction`, `nameORgeneMolecularInteraction`, and `nameORgeneMolecularInteraction` coming from?

markquintontulloch: @oblodgett The 'OR' is a separator for the fields in order to get the searchDAO query to perform an OR search, so the fields are: 
geneMolecularInteraction.interactionId
geneMolecularInteraction.aggregationDatabase.name
geneMolecularInteraction.interactionSource
geneMolecularInteraction.crossReferences.displayName

oblodgett: Ok... I don't remember it working this way... let me take a look

oblodgett: Ok seeing the code:
```java
String[] filterNames = filterName.split("OR");
	Arrays.stream(filterNames).forEach(indivFilterName -> {
	BoolQueryBuilder orClause = getBooleanAndedQueryBuilder(indivFilterName.trim(), filterValue);
	outerAndClause.should(orClause);
});
```
Weird... I don't remember us doing this but ok.

oblodgett (approved):

PR #1300, opened 2025-03-31
Stop escaping characters in OR search

markquintontulloch:


agr_api/src/main/java/org/alliancegenome/api/service/ESService.java
```diff
@@ -95,7 +95,7 @@ private void generateFilter(BoolQueryBuilder bool, String filterName, String fil
 			//Log.info("Or Filter: " + filterName + " " + filterValue);
 			BoolQueryBuilder orClause = boolQuery();
 			String[] elements = filterValue.split("\\|");
-			Arrays.stream(elements).forEach(element -> orClause.should(QueryBuilders.termQuery(filterName, escapeValue(element))));
+			Arrays.stream(elements).forEach(element -> orClause.should(QueryBuilders.termQuery(filterName, element)));
```
oblodgett, line 98: @cmpich and I had a conversation about this long ago, and I thought it would come back to bite us... maybe this is it? I think one of the variants pages might break given that we are looking up names via hgvsgName.

`/GeneRESTInterface.java:		@QueryParam("filter.hgvsgName") String hgvsgName,`
`AlleleVariantSequenceFiltering.java:		filterFieldMap.put(FieldFilter.VARIANT_HGVS_G, hgvsgNameFilter);`

P.S. I am in favor of NOT escaping the input.

markquintontulloch: Escaping the characters breaks the OR keyword searches if the terms contain brackets.  We don't escape the characters for single term searches so I don't see the need to do it for OR

oblodgett (approved):

PR #1301, opened 2025-04-01
Fix OR search pt 2 - prevent string escaping in helper method too

markquintontulloch:


oblodgett (approved):

markquintontulloch:
@oblodgett Not entirely sure why we have this helper class - appears to just be duplicating methods in ESService

oblodgett:
Yes I think the code was copied and pasted at one point and the orignal was not deleted / refactored... @cmpich?

PR #1339, opened 2025-06-18
Scrum 5078 Refactor

oblodgett:


FB-ping (approved):
@oblodgett thanks, seems you re-structure the data so it will look like other data. anyway,  I tested it on local and all seems fine. thanks.

PR #1344, opened 2025-06-24
Add mode DEBUG INFO. Will be removed later.

cmpich:


markquintontulloch (approved):

oblodgett (commented):
@cmpich there should be no difference between running this locally and in stage

PR #1357, opened 2025-07-10
SCRUM-4989 Update Xenopus assembly versions

markquintontulloch:


oblodgett (commented):
I assume there is loader work going on for this also?

oblodgett (approved):

PR #1358, opened 2025-07-10
change in accession caching

VarunReddy1111:


agr_cacher/src/main/java/org/alliancegenome/cacher/cachers/SiteMapCacher.java
```diff
@@ -66,11 +68,39 @@ protected void cache() {
...
 	private void cacheAccession(String type, List<String> keyList, CacheAlliance cache) {
-
+		HashMap<String, String> batchMap = new HashMap<>();
+		int totalProcessed = 0;
+		int batchCount = 0;
+		
+		log.info("Starting batched bulk caching for " + type + " with " + keyList.size() + " keys");
+		
 		for (String key : keyList) {
-			cacheService.putCacheEntry(key, "https://www.alliancegenome.org/" + type + "/" + key, View.Default.class, cache);
```
oblodgett, line 71: I would suggest wrapping the original for loop in a try cache to see what the error is.

VarunReddy1111: Tried doing that, was not throwing any error, it is either getting stuck for hours or successfully finished without data being cached

oblodgett: @VarunReddy1111 there has got to be something causing this... I will be back tomorrow and will take a look at it. The other option is to put a ProcessDisplayHelper on the for loop and set the display time to like 1 second and see what happens.

oblodgett:
@VarunReddy1111 do we still need this PR open?

VarunReddy1111:
Closing this as it is no more required

PR #1386, opened 2025-09-24
run GeneToGeneOrthologyIndexer not in parallel mode for now

cmpich:


oblodgett (commented):
@cmpich this is going to add about 3.5 hours to the indexer... it could be that both these cannot run at the same time, and or also we may need to up the box size for stage-curation.

cmpich:
We do need a good run so other work can continue. While that is going on we can experiment with other solutions. 
Before you change of running indexer in parallel we had about 10 hours run successful. Our current solution would be better than that as all indexer ran in series. 
I am happy to experiment then with a bigger box.

PR #1397, opened 2025-10-07
remove association api calls from AlleleSummaryCurationIndexer

cmpich:


adamgibs (approved):

oblodgett (commented):
can the interfaces be removed also?

PR #1398, opened 2025-10-08
remove parallelization

cmpich:


agr_indexer/src/main/java/org/alliancegenome/indexer/Main.java
```diff
@@ -48,7 +48,9 @@ public static void main(String[] args) {
 				Indexer i = (Indexer) ic.getIndexClazz().getDeclaredConstructor(IndexerConfig.class).newInstance(ic);
 				indexers.put(ic.getTypeName(), i);
 				if (ic.getRunInParallel()) {
-					parallelMap.put(ic.getTypeName(), i);
+/// TODO: stopped using parallel as it caused slowdown
+// parallelMap.put(ic.getTypeName(), i);
+					sequentialMap.put(ic.getTypeName(), i);
```
oblodgett: instead of switching this i would suggest just setting the fixedThreadPool to 1 this way at least the neo indexers and curation indexes can run in parallel

agr_indexer/src/main/java/org/alliancegenome/indexer/config/IndexerConfig.java
```diff
@@ -45,7 +45,7 @@ public enum IndexerConfig {
 	AlleleSummaryIndexer("alleleSummary", AlleleSummaryCurationIndexer.class, 4, 500, 500, 4, 1, true),
```
adamgibs: Should we increase this batch size as well?

adamgibs (approved):

PR #1418, opened 2025-10-30
do not run any indexer in parallel

cmpich:


oblodgett (commented):
@cmpich is there an issue? I would guess this is not going to fix it.

cmpich:
The issue is that the indexer is always failing when the GeneMolecularInteractionCurationIndexer is running while the Neo4j indexer are running. I don't have a better solution for now. I want to make sure it's not related to concurrent Neo4j access. TNeo4j is always getting unresponsive. maybe a memory issue?

oblodgett (approved):
@cmpich I don't think this will fix it, but anyway

PR #1422, opened 2025-11-08
Scrm 4072

cmpich:


agr_java_core/src/main/java/org/alliancegenome/neo4j/entity/SpeciesType.java
```diff
@@ -175,4 +175,13 @@ public static HashMap<String, Integer> getSpeciesOrderByTaxonID(String taxonId)
 		return map;
 	}
 
+	public static int getPhylogeneticSortOrder(String taxonID) {
```
oblodgett, line 178: @cmpich this order should be on the taxon object... if not we need to add it, so that we can get rid of this Enum and have it in the database.

PR #1430, opened 2025-11-18
Scrum 5363

cmpich:


agr_indexer/src/main/java/org/alliancegenome/indexer/config/IndexerConfig.java
```diff
@@ -35,6 +27,7 @@ public enum IndexerConfig {
 	LiteratureIndexer("literature", LiteratureIndexer.class, 4, 5000, 5000, 1, 1, true),
 	
 	// Curation Indexers
+	VariantSummaryIndexer("variantSummary", org.alliancegenome.indexer.indexers.VariantSummaryIndexer.class, 1, 3000, 1500, 8, 1, true),
```
VarunReddy1111: This can be shortened to just indexer class name

VarunReddy1111 (approved):

oblodgett (commented):
i would put the pages in a queue so that this indexer could be multi thread

PR #1431, opened 2025-11-20
SCRUM-5602 remove constraint that checks if entities of DAs are found…

cmpich:
… in Neo


agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
```diff
@@ -93,8 +72,7 @@ protected void index() {
 		alleleService = new AlleleDiseaseAnnotationService();
 		agmService = new AGMDiseaseAnnotationService();
 		vocabTermService = new VocabularyTermService();
-		diseaseRepository = new DiseaseRepository();
-		
+		DiseaseRepository diseaseRepository = new DiseaseRepository();
```
oblodgett, line 75: @cmpich this has to be closed why create a new one here?

oblodgett (changes requested):

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/DiseaseAnnotationCurationIndexer.java
cmpich: I only changed the scope of this object, from global to method-scoped. It was created on the class level before.

oblodgett: So not understanding why the scope on this needs to be changed in the first place. Plus any "Repository" depends on neo so this doesn't remove the dependency on neo at all.

cmpich: I changed the scope as the diseaseRepository instance is only used in the index() method. Ture, there is still the dependency on Neo but only that method was using this object. 
In the next step we are going to remove this and replace it with a curation method. For now, the closure is still being calculated through the neo4j repository.

cmpich: The repository gets closed at the end of the index() method, nothing has really changed except the scope. In general, you define your variables with the smallest scope necessary to fulfill its purpose.

oblodgett (approved):

PR #1433, opened 2025-11-26
Changed ExpressionRibbonESService to handle race condition using sync…

VarunReddy1111:
…ronized block

### DIFF
--- a/agr_api/src/main/java/org/alliancegenome/api/service/ExpressionRibbonESService.java
+++ b/agr_api/src/main/java/org/alliancegenome/api/service/ExpressionRibbonESService.java
@@ -45,7 +45,7 @@ public class ExpressionRibbonESService extends ESService {
 
 	public static final String UNDEFINED = "undefined";
 
-	private static RibbonSummary ribbonSummary;
+	private static volatile RibbonSummary ribbonSummary;
 
 	public RibbonSummary getRibbonSectionInfo() {
 		// get a deep clone of a template object
@@ -120,8 +120,16 @@ public class ExpressionRibbonESService extends ESService {
 		Pagination pagination = new Pagination(1, 250000, null, null);
 		JsonResultResponse<GeneExpressionDocument> expressionAnnotations = expressionService.getExpressionAnnotations(List.of(geneID), null, null, pagination);
 
-		Gene gene = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation().getExpressionAnnotationSubject();
-		String dataProvider = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation().getDataProvider().getAbbreviation();
+		String dataProvider;
+		Gene gene;
+		if (CollectionUtils.isNotEmpty(expressionAnnotations.getResults())) {
+			gene = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation().getExpressionAnnotationSubject();
+			dataProvider = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation().getDataProvider().getAbbreviation();
+		} else {
+			gene = new Gene();
+			dataProvider = null;
+		}
+
 		RibbonEntity entity = new RibbonEntity();
 		entity.setId(geneID);
 		entity.setLabel(gene.getGeneSymbol().getDisplayText());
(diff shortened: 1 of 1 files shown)


agr_api/src/main/java/org/alliancegenome/api/service/ExpressionRibbonESService.java
```diff
@@ -120,8 +120,16 @@ private RibbonEntity getExpressionRibbonSummary(String geneID) {
...
-		Gene gene = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation().getExpressionAnnotationSubject();
-		String dataProvider = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation().getDataProvider().getAbbreviation();
+		String dataProvider;
+		Gene gene;
+		if (CollectionUtils.isNotEmpty(expressionAnnotations.getResults())) {
+			gene = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation().getExpressionAnnotationSubject();
+			dataProvider = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation().getDataProvider().getAbbreviation();
+		} else {
+			gene = new Gene();
+			dataProvider = null;
```
cmpich, line 131: combine this assignment with line 123
String dataProvider = null;
then you don't need another dataProvider assignment in the else clause.
Same with the gene assignment. You can get rid of the else clause all together.

agr_api/src/main/java/org/alliancegenome/api/service/ExpressionRibbonESService.java
```diff
@@ -120,8 +120,16 @@ private RibbonEntity getExpressionRibbonSummary(String geneID) {
 		Pagination pagination = new Pagination(1, 250000, null, null);
 		JsonResultResponse<GeneExpressionDocument> expressionAnnotations = expressionService.getExpressionAnnotations(List.of(geneID), null, null, pagination);
 
-		Gene gene = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation().getExpressionAnnotationSubject();
-		String dataProvider = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation().getDataProvider().getAbbreviation();
+		String dataProvider;
+		Gene gene;
+		if (CollectionUtils.isNotEmpty(expressionAnnotations.getResults())) {
+			gene = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation().getExpressionAnnotationSubject();
```
cmpich: I would define an annotation object used in line 126 and 127
GeneExpressionAnnotation annotation = expressionAnnotations.getResults().get(0).getGeneExpressionAnnotation();
the two assignments then get more readable.

agr_api/src/main/java/org/alliancegenome/api/service/ExpressionRibbonESService.java
```diff
@@ -253,59 +261,66 @@ public RibbonSummary getRibbonSections() {
...
+
+				SectionSlim allSlimElement = new SectionSlim();
+				allSlimElement.setId(id);
+				allSlimElement.setLabel(names.get(1));
+				allSlimElement.setTypeAll();
+				section.addDiseaseSlim(allSlimElement);
+				tempRibbonSummary.addRibbonSection(section);
+
+				List<GOTerm> goSlimList = slimTerms.getGoSlimTerms();
+				if (id.equals(ExpressionCacheRepository.GO_CC_ROOT)) {
```
cmpich: You could use a switch command series making the logic a bit more readable:
			switch (id) {
				case ExpressionCacheRepository.GO_CC_ROOT:
					goSlimList.forEach(term -> {
						if (term.getCurie().equals(ExpressionCacheRepository.GO_CC_ROOT)) {
							section.setDescription(term.getDefinition());
							allSlimElement.setDescription(term.getDefinition());
						} else {
							SectionSlim slim = getSectionSlim(term.getCurie(), term.getName(), term.getDefinition());
							section.addDiseaseSlim(slim);
						}
					});
					break;
...

agr_api/src/main/java/org/alliancegenome/api/service/ExpressionRibbonESService.java
```diff
@@ -253,59 +261,66 @@ public RibbonSummary getRibbonSections() {
...
-				goSlimList.forEach(term -> {
-					if (term.getCurie().equals(ExpressionCacheRepository.GO_CC_ROOT)) {
-						section.setDescription(term.getDefinition());
-						allSlimElement.setDescription(term.getDefinition());
-					} else {
-						SectionSlim slim = getSectionSlim(term.getCurie(), term.getName(), term.getDefinition());
-						section.addDiseaseSlim(slim);
-					}
-				});
+		synchronized (ExpressionRibbonESService.class) {
```
oblodgett, line 265: @VarunReddy1111 you shouldn't need thread safe stuff here, are multiple threads in this code at the same time?

VarunReddy1111: Yes, there would be multiple threads for different requests which read/write this static variable ribbonSummary. Hence, making it thread safe

agr_api/src/main/java/org/alliancegenome/api/service/ExpressionRibbonESService.java
VarunReddy1111: It can changed for gene, but for data provider, the code expects it to initialized once and not changed as this variable is accessed in below lambda function

cmpich (approved):

agr_api/src/main/java/org/alliancegenome/api/service/ExpressionRibbonESService.java
oblodgett: ribbon summary should not be static it should be a new instance for each request

VarunReddy1111: ribbon summary contains sections and entities. The section part remains same for all request irrespective of any gene and entities change accordingly with each gene. Here we create a static ribbonSummary variable with the sections info and once the sections are captured, it creates a local ribbonSummary variable and stores into that without changing the static variable and now this local ribbonSummary variable gets modified to add the entities for each gene and this local ribbonSummary would be returned.

VarunReddy1111: Thus, instead of creating new sections for each request, we create it once and store in a static variable which is accessed by incoming requests

PR #1435, opened 2025-12-03
Changes to expression ribbon service

VarunReddy1111:


agr_api/src/main/java/org/alliancegenome/api/service/ExpressionRibbonESService.java
```diff
@@ -106,6 +107,26 @@ public GeneExpressionRibbonSummaryDocument getGeneExpressionRibbonSlimTerms() {
 		return null;
 	}
 
+
+	public Map<String, Object> getGene(String geneID) {
+		BoolQueryBuilder boolQuery = boolQuery();
+		boolQuery.filter(new TermQueryBuilder("category", "gene"));
```
cmpich, line 114: Could you add a comment here that the category=gene may change when we convert the search functionality?

VarunReddy1111: Sure

adamgibs (approved):

PR #1495, opened 2026-02-26
SCRUM-5805 update API to accomodate the go_search_result doc

adamgibs:


agr_api/src/main/java/org/alliancegenome/api/service/SearchService.java
```diff
@@ -319,7 +328,7 @@ public void addRelatedDataLinks(Map<String, Object> result) {
 		if (StringUtils.equals(category, "gene")) {
 			links.add(getRelatedDataLink("disease", "genes", nameKey));
 			links.add(getRelatedDataLink("allele", "genes", nameKey));
-			links.add(getRelatedDataLink("go", "go_genes", nameKey));
+			links.add(getRelatedDataLink(Category.GO.getName(), "genes", nameKey));
```
oblodgett, line 331: Given that this is a change I would rather see all of them switched out not just the one

oblodgett (approved):

PR #1499, opened 2026-02-27
Scrum 5778a

cmpich:


agr_api/src/main/java/org/alliancegenome/api/service/helper/SearchHelper.java
```diff
@@ -350,6 +366,50 @@ public ArrayList<AggResult> formatAggResults(String category, SearchResponse res
...
 	}
 
+	/**
+	 * Merge the "allele" and "variant_search_result" aggregation buckets into a single "allele_variant" bucket.
+	 */
+	private void mergeAlleleVariantBuckets(AggResult aggResult) {
+		long combinedCount = 0;
+		List<AggDocCount> toRemove = new ArrayList<>();
+		for (AggDocCount bucket : aggResult.getValues()) {
+			if ("allele".equals(bucket.getKey()) || "variant_search_result".equals(bucket.getKey())) {
```
oblodgett: Can these texts be switched over, to the Category.ALLELE etc.

agr_api/src/main/java/org/alliancegenome/api/service/SearchService.java
```diff
@@ -243,7 +244,11 @@ public BoolQueryBuilder buildQuery(String queryTerm, String category, Multivalue
 
 		// apply filters if a category has been set
 		if (StringUtils.isNotEmpty(category)) {
-			bool.filter(new TermQueryBuilder("category", category));
+			if (Category.ALLELE_VARIANT.getName().equals(category)) {
+				bool.filter(termsQuery("category", "allele", "variant_search_result"));
```
oblodgett: I would think these filters to be the categories as well?

agr_api/src/main/java/org/alliancegenome/api/service/helper/SearchHelper.java
```diff
@@ -335,7 +345,13 @@ public ArrayList<AggResult> formatAggResults(String category, SearchResponse res
 
 		if (category == null) {
 			Terms aggs = res.getAggregations().get("categories");
-			AggResult ares = new AggResult("category", aggs, categoryFilters.keySet());
+			// Allow allele and variant_search_result through for merging
+			Set<String> acceptableKeys = new HashSet<>(categoryFilters.keySet());
+			acceptableKeys.add("allele");
```
oblodgett: Should the category enum be used here also?

oblodgett (approved):

PR #1501, opened 2026-03-02
SCRUM-5778 enable the AlleleIndexer again to reinstantiate the allele…

cmpich:
… search
Upgrade curation library to 0.46.27


oblodgett:
@cmpich I thought the allele search result's were in place and we didn't need this?

cmpich:
the allele search results are not in place as far as I can tell. 
I am pretty close in doing it but I'd like to not include this in the current release.

oblodgett:
@cmpich we have to... allele's and variant's are all or nothing at this point.

oblodgett (approved):

PR #1506, opened 2026-03-05
SCRUM-5821 Migrate HTP dataset indexing from Neo4j to curation API

adamgibs:
Replace DatasetIndexer with HTPDatasetSearchResultCurationIndexer following the GO search result pattern. Category changes from "dataset" to "htp_dataset_search_result".

Also standardize categoryFilters in SearchHelper to use Category enum instead of hardcoded strings.

Bump curation.version to v0.46.32.

### DIFF
--- a/agr_api/src/main/java/org/alliancegenome/api/service/helper/SearchHelper.java
+++ b/agr_api/src/main/java/org/alliancegenome/api/service/helper/SearchHelper.java
@@ -26,7 +26,7 @@ public class SearchHelper {
 
 	private HashMap<String, List<String>> categoryFilters = new HashMap<>() {
 		{
-			put("gene", new ArrayList<>() {
+			put(Category.GENE.getName(), new ArrayList<>() {
 				{
 					add("species");
 					add("biotypes");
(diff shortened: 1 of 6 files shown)


agr_api/src/main/java/org/alliancegenome/api/service/helper/SearchHelper.java
```diff
@@ -75,15 +75,15 @@ public class SearchHelper {
...
+			put(Category.MODEL.getName(), new ArrayList<>() {
 				{
 					add("species");
 					add("diseasesAgrSlim");
 					add("genes");
 					add("alleles");
 				}
 			});
-			put("variant_search_result", new ArrayList<>() {
+			put(Category.VARIANT_SEARCH_RESULT.getName(), new ArrayList<>() {
```
oblodgett: Can you refactor this back to be just `Category.VARIANT` that way in the future we can just switch from "go" to "go_search_result" like you did and the code doesn't need to know the actual document category name.

cmpich (approved):

PR #1507, opened 2026-03-06
Fixing the download endpoint to have HTP variants

VarunReddy1111:


agr_api/src/main/java/org/alliancegenome/api/controller/GeneController.java
```diff
@@ -113,7 +113,7 @@ public JsonResultResponse<ESDocument> getAllelesPerGene(String id, Integer limit
 
 		long startTime = System.currentTimeMillis();
 		Pagination pagination = new Pagination(page, limit, sortBy, asc);
-		pagination.addFilterOption("allele.alleleSymbol.displayText", symbol);
+		pagination.addFilterOption("allele.alleleSymbol.displayTextORvariants.curatedVariantGenomicLocations.hgvs", symbol != null ? symbol.strip() : null);
```
oblodgett: Symbol is coming in from the browser? I would suggest we take the value as it is. If there is an issue with extra spaces deal with that on the UI side.

agr_java_core/src/main/java/org/alliancegenome/core/translators/tdf/AlleleToTdfTranslator.java
```diff
@@ -44,56 +46,78 @@ public String getAllRows(List<ESDocument> annotations) {
...
+					return List.of(getBaseDownloadRow(doc, null));
 				}
 			})
 			.flatMap(Collection::stream)
 			.collect(Collectors.toList());
 	}
 
+	private List<Variant> getVariants(AVSParentDocument doc) {
+		if (doc instanceof AlleleSummaryDocument asd) {
+			return asd.getVariants();
```
cmpich: You could use the new switch statement instead of the if-clauses and instanceof check.

oblodgett (approved):

PR #1521, opened 2026-03-13
SCRUM-4816: Add GeneSearchResultCurationIndexer with ID-batch pattern

oblodgett:
## Summary
- Add GeneSearchResultCurationIndexer using ID-batch pattern: fetches all gene IDs, partitions into batches of 1000, 4 threads call `findSearchResultByIds` endpoint
- Indexer is currently commented out in IndexerConfig pending curation API release with the new endpoint
- pom.xml curation.version set to 0.0.0 (local install) -- needs update to actual release version after curation PR merge

## Test plan
- [ ] After curation release: update `curation.version` in pom.xml to released version
- [ ] Uncomment GeneSearchResultCurationIndexer in IndexerConfig
- [ ] Run indexer end-to-end, verify gene documents are indexed correctly


oblodgett:
Merged into fix_variant_indexer branch (PR #1522) to combine all three PRs into one.

PR #1524, opened 2026-03-16
Allele search result indexer

cmpich:


oblodgett (approved):
Love it, to see code removed that is.

PR #1530, opened 2026-03-22
SCRUM-5829 make MNV variants and all other allelic variants show up i…

cmpich:
…n allele search result


agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleSearchResultConverter.java
```diff
@@ -56,7 +57,14 @@ public List<AlleleSearchResultDocument> convert(List<AlleleSummaryDocument> alle
 			}
 
 			if (doc.getAlleleOfGene() != null && doc.getAlleleOfGene().getGeneSymbol() != null) {
-				searchDoc.setGenes(List.of(doc.getAlleleOfGene().getGeneSymbol().getDisplayText()));
+				String geneSymbol = doc.getAlleleOfGene().getGeneSymbol().getDisplayText();
+				if (allele.getTaxon() != null) {
+					SpeciesType speciesType = SpeciesType.getTypeByNameField(allele.getTaxon().getName());
```
oblodgett: Please don't use this class. We have this same info on the biological entity class biologicalentity.taxon.species in this specific case allele.getTaxon().getSpecies().getName() the species also has the getAbbreviation() method.

oblodgett (changes requested):

agr_java_core/src/main/java/org/alliancegenome/es/model/AlleleSearchResultDocument.java
```diff
@@ -28,6 +28,7 @@ public AlleleSearchResultDocument() {
 	private String species;
 	private Double popularity;
 	private String alterationType;
+	private List<String> variantType;
```
oblodgett: Does the search require this to be the name? I would assume it should be called variantTypes?

oblodgett (approved):

PR #1537, opened 2026-03-26
SCRUM-5829 collate both SGD spcies into a single species facet

cmpich:


agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleSearchResultConverter.java
```diff
@@ -39,9 +39,10 @@ public List<AlleleSearchResultDocument> convert(List<AlleleSummaryDocument> alle
 
 			NCBITaxonTerm taxon = allele.getTaxon();
 			if (taxon != null) {
-				searchDoc.setSpecies(taxon.getName());
+				String speciesName = normalizeSpeciesName(taxon.getName());
```
oblodgett, line 42: speciesName needs to be allele.getTaxon().getSpecies().getName() and we need to revert the SGD special handling of the name.

oblodgett (changes requested):

agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/AlleleSearchResultConverter.java
oblodgett: We are going to need to do this story: https://agr-jira.atlassian.net/browse/SCRUM-2758 to get rid of this issue.

PR #1538, opened 2026-03-26
SCRUM-5829 get variant type and moelcular consequence se the same jso…

cmpich:
…n attribute names as the HTP variants

### DIFF
--- a/agr_java_core/src/main/java/org/alliancegenome/es/model/AlleleSearchResultDocument.java
+++ b/agr_java_core/src/main/java/org/alliancegenome/es/model/AlleleSearchResultDocument.java
@@ -29,10 +29,8 @@ public class AlleleSearchResultDocument extends ESDocument {
 	private String species;
 	private Double popularity;
 	private String alterationType;
-	@JsonProperty("variantType")
-	private List<String> variantTypes;
-	@JsonProperty("molecularConsequence")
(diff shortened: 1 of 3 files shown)


agr_java_core/src/main/java/org/alliancegenome/es/model/AlleleSearchResultDocument.java
```diff
@@ -29,7 +29,9 @@ public AlleleSearchResultDocument() {
 	private String species;
 	private Double popularity;
 	private String alterationType;
+	@JsonProperty("variantType")
```
oblodgett: do these HAVE to be the names?

cmpich:
This also includes to not have count numbers in ES truncated at 10,000.

agr_java_core/src/main/java/org/alliancegenome/es/model/AlleleSearchResultDocument.java
oblodgett: instead of remapping think it would be better to just change the names in java

cmpich: We could change the field name as well to match the VariantSearchResultDocument notation.

oblodgett (approved):

PR #1557, opened 2026-04-07
SCRUM-5881 add variants as attributes on allele document

cmpich:


agr_api/src/main/java/org/alliancegenome/api/service/helper/SearchHelper.java
```diff
@@ -216,6 +216,8 @@ public List<String> getSearchFields() {
 			add("variants");
 			add("variants.keyword");
 			add("variants.standardText");
+			add("variantHgvs");
+			add("variantHgvs.keyword");
```
oblodgett, line 220: These should be added to the variantSynonyms field and not add an extra field.

oblodgett (changes requested):

PR #1560, opened 2026-04-09
SCRUM-5829 generate variant summary docs for alleles

cmpich:


agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/AlleleSummaryCurationIndexer.java
```diff
@@ -79,6 +82,7 @@ protected void startSingleThread(LinkedBlockingDeque<String> queue) {
 				// Convert to derived documents first (consumes transport-only fields)
 				List<SequenceSummaryDocument> sequenceDocs = sequenceSummaryConverter.convert(response.getResults());
 				List<AlleleSearchResultDocument> searchDocs = alleleSearchResultConverter.convert(response.getResults());
+				List<VariantSummaryDocument> variantDocs = variantSummaryConverter.convert(response.getResults());
```
oblodgett, line 85: I think this causes double entries in the Allele table on the gene page but we'll see.

oblodgett (approved):

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/AlleleSummaryCurationIndexer.java
cmpich: We'll see.

oblodgett (approved):

PR #1576, opened 2026-04-20
Reorged variant code to prevent single task issues

oblodgett:


agr_variant_indexer/src/main/java/org/alliancegenome/indexer/variant/es/managers/BulkSubmitter.java
```diff
@@ -0,0 +1,67 @@
...
+			}
+			return BulkSubmitResult.success();
+
+		} catch (ElasticsearchStatusException e) {
+			return BulkSubmitResult.fullFailure("HTTP " + e.status().getStatus() + ": " + e.getMessage(), false, true);
+		} catch (ConnectTimeoutException e) {
+			return BulkSubmitResult.fullFailure("ConnectTimeoutException: " + e.getMessage(), true, false);
+		} catch (SocketTimeoutException e) {
+			return BulkSubmitResult.fullFailure("SocketTimeoutException: " + e.getMessage(), true, false);
+		} catch (ConnectionClosedException e) {
```
cmpich, line 57: Do we need to handle them all separately? I think we could combine them in a single catch clause and display the message of the exception. The type of exception should be clear from the error message and the stack trace.

oblodgett: Yes we could combine them, however I am trying to explicitly see which errors are getting thrown and why. Plus there different options with the flags for the different types of errors.

PR #1582, opened 2026-04-28
SCRUM-6006 add gene systematic name to HTP variants

cmpich:


agr_java_core/src/main/java/org/alliancegenome/core/variant/converters/VariantSummaryConverter.java
```diff
@@ -394,6 +394,9 @@ private Pair<List<PredictedVariantConsequence>, HashSet<String>> getConsequences
 								if (cachedGene.getGeneSynonyms() != null) {
 									gene.setGeneSynonyms(cachedGene.getGeneSynonyms());
 								}
+								if (cachedGene.getGeneSystematicName() != null) {
+									gene.setGeneSystematicName(cachedGene.getGeneSystematicName());
+								}
```
oblodgett, line 399: @cmpich can you add this to the Variant Search Result also, not sure if that document is using the gene directly or creating a copy.

cmpich: The VariantSearchResult object has a collection of systematicNames (strings) and that gets populated in this class. I am not sure what you are referring to.

VarunReddy1111 (approved):

PR #1589, opened 2026-05-06
SCRUM-5126 retire Neo4j closure call in DiseaseAnnotationCurationIndexer

adamgibs:
Replace DiseaseRepository.getDOClosureChildMapping() with a new DoTermService that builds the same Map<String, Set<String>> via the curation API's /ontologytermclosure/document endpoints.

DoTermService pages through getAllIds + findByIds in batches and adds a self-edge per key to mirror the original Neo4j IS_A_PART_OF_CLOSURE semantics (ontologytermclosure has no self rows). Failures fall through to ExceptionCatcher.report + System.exit(-1).


oblodgett:
The partition method: I know claude probably generated this method, but I feel like I have seen it before... and we should probably pull it out into its own static method somewhere.

adamgibs:
@oblodgett Yeah, it's a pure duplicate of the one in the Indexer class. I decided to keep it in as apposed to extending the Indexer class with a service class or changing the method to public. I can pull them out into a util class, but it'll mean all of the indexers imports need to be updated.

oblodgett:
Ya... under agr_java_core there is a util package. I think we should have a ListUtils class then you can just call ListUtils.partition(), imports are no big deal. Given claude just tell him the places you want the method removed from and it will do the refactor for you.

adamgibs:
@oblodgett Method has been extracted

oblodgett:
something doesn't look right in your PR. Those files shouldn't be "new" they should be existing... did you branch from the correct base?

oblodgett:
I see what is happening: target is not gitignored and it should be. If you could add that to the gitignore file that would be great. I should have caught that in my initial PR.

adamgibs:
@oblodgett file generator target file added to gitignore

oblodgett:
But they are still in the commit.

oblodgett (approved):

PR #1595, opened 2026-05-07
SCRUM-6035

cmpich:


agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/AuditedObjectIndexerMixin.java
```diff
@@ -0,0 +1,24 @@
...
+ * Jackson MixIn applied to AuditedObject during ES indexing only. SCRUM-6035: drop curation-internal
+ * fields (internal, dbDateCreated, dbDateUpdated) from every public ES document — they leak via nested
+ * entities (Gene, Allele, Variant, etc.) because most indexers serialize without an explicit @JsonView.
+ * Adding these field-level @JsonIgnore overrides via a MixIn suppresses them globally on the indexer's
+ * ObjectMapper without modifying agr_curation entity annotations.
+ */
+public abstract class AuditedObjectIndexerMixin {
+
+	@JsonIgnore
+	Boolean internal;
```
oblodgett, line 17: @cmpich do we want obsolete also? I think for now Chris said that there would not be any, given that all our queries are where obsolete = false;

oblodgett (approved):

PR #1596, opened 2026-05-07
SCRUM-6042 invoke VariantSummaryDocument.removeTransportFields in variant indexers

cmpich:
## Summary

- `VariantSummaryDocument.removeTransportFields()` was added in `agr_curation_api` v0.48.16 (`agr_curation` PR for SCRUM-6042) and the version bump landed on `stage` in #1587, but the method is never called — dead code.
- This PR wires it into the two indexer paths that produce `VariantSummaryDocument`s, so the cross-references suppression actually takes effect.

## Why

Without these calls, the indexed `_source` for variant-summary docs continues to carry the per-transcript `transcriptGeneAssociations[*].transcriptGeneAssociationObject.crossReferences` lists. For TP53 (HGNC:11998) that means ~1,200 cross-reference entries per variant in the indexed JSON; on a 10-variant page the response is ~1.3 MB. The follow-up `removeTransportFields()` strips those, taking it to roughly 50–100 KB.

## Files

- `agr_indexer/.../curation/VariantSummaryCurationIndexer.java:74` — direct path (TP53 and other variant-only docs).
- `agr_indexer/.../curation/AlleleSummaryCurationIndexer.java:90` — derived path for `variantDocs` produced by the converter.

## Test plan

- [ ] Run `VariantSummaryCurationIndexer` against stage curation API; confirm reindex completes without errors.
- [ ] Curl `/api/gene/HGNC:11998/alleles?page=1&limit=10` on stage after deploy; expect transcript-gene `crossReferences` lists absent and total payload ~50–100 KB (down from ~1.3 MB).
- [ ] Spot-check an allele with variants (e.g. `/api/allele/ZFIN%3AZDB-ALT-990423-22`) to confirm AlleleSummaryDocument-derived variant docs are also slimmed.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/VariantSummaryCurationIndexer.java
```diff
@@ -72,6 +72,7 @@ protected void startSingleThread(LinkedBlockingDeque<String> queue) {
 				String page = queue.takeFirst();
 				
 				SearchResponse<VariantSummaryDocument> response = variantApi.findDocuments(Integer.valueOf(page), indexerConfig.getBufferSize(), params);
+				response.getResults().forEach(VariantSummaryDocument::removeTransportFields);
```
oblodgett, line 75: We should just drop the json views here vs doing this?

agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/curation/AlleleSummaryCurationIndexer.java
```diff
@@ -87,6 +87,7 @@ protected void startSingleThread(LinkedBlockingDeque<String> queue) {
 
 				// Strip fields only needed by derived documents before indexing to ES
 				response.getResults().forEach(AlleleSummaryDocument::removeTransportFields);
+				variantDocs.forEach(VariantSummaryDocument::removeTransportFields);
```
oblodgett, line 90: These fields were not converted in the variantSummaryConverter.convert

PR #1604, opened 2026-05-15
SCRUM-5124 add AGM search indexer

cmpich:


oblodgett (approved):

agr_indexer/src/main/java/org/alliancegenome/indexer/config/IndexerConfig.java
oblodgett: Can we turn this one off? If we want to compare we have the index in production.

cmpich: Sure, makes sense.

PR #1610, opened 2026-05-19
KANBAN-1322 add disease ontology browser API endpoints

jdepons:
## Summary
New endpoints powering the disease ontology browser UI (matching `feature/KANBAN-1322` in agr_ui):

- `GET /api/disease/batch?ids=...` - batch fetch `DiseaseSummaryDocument` for many curies in one ES round-trip
- `GET /api/disease/counts?ids=...` - gene/model/allele annotation counts per disease curie
- `GET /api/disease/{id}/ancestors` - root-first ancestor chain (capped at depth 30, cycle-safe)

Both batch endpoints reject requests with more than 500 ids (`BadRequestException`) so a large `ids` list cannot blow up an ES query.

## Test plan
- [ ] `curl 'http://stage.alliancegenome.org/api/disease/batch?ids=DOID:14330,DOID:1287'` returns a map of curie -> document
- [ ] `curl 'http://stage.alliancegenome.org/api/disease/counts?ids=DOID:14330,DOID:1287'` returns `{ DOID:14330: { genes, models, alleles }, ... }`
- [ ] `curl 'http://stage.alliancegenome.org/api/disease/DOID:14330/ancestors'` returns the ancestor chain root-first
- [ ] Sending 501 comma-separated ids to either batch endpoint returns HTTP 400
- [ ] Swagger UI at `/q/swagger-ui/` lists the new endpoints

### DIFF
--- a/agr_api/src/main/java/org/alliancegenome/api/entity/DiseaseTermStub.java
+++ b/agr_api/src/main/java/org/alliancegenome/api/entity/DiseaseTermStub.java
@@ -1,0 +1,31 @@
+package org.alliancegenome.api.entity;
+
+public class DiseaseTermStub {
+
+	private String curie;
+	private String name;
+
+	public DiseaseTermStub() {
+	}
+
+	public DiseaseTermStub(String curie, String name) {
+		this.curie = curie;
+		this.name = name;
+	}
+
+	public String getCurie() {
+		return curie;
+	}
+
+	public void setCurie(String curie) {
+		this.curie = curie;
+	}
+
+	public String getName() {
+		return name;
+	}
+
+	public void setName(String name) {
+		this.name = name;
+	}
+}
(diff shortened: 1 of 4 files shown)


agr_api/src/main/java/org/alliancegenome/api/entity/DiseaseTermStub.java
```diff
@@ -0,0 +1,31 @@
+package org.alliancegenome.api.entity;
+
+public class DiseaseTermStub {
```
oblodgett, line 3: @jdepons a better way to do this, which is what we typically is create a JsonView that you put on the interface @JsonView{PublicView.DiseaseClosterView.class) and then you annotate the document with that class. Then you don't need a stub class and you can just return the document list and all that will come out via json is objects with that "view".

jdepons:
Closing temporarily — will reopen once the Quarkus 3.29.4 / resteasy issue on stage is resolved.

PR #1612, opened 2026-05-19
KANBAN-1322 add disease ontology browser API endpoints

jdepons:
## Summary
New endpoints powering the disease ontology browser UI (matching `feature/KANBAN-1322` in agr_ui):

- `GET /api/disease/batch?ids=...` - batch fetch `DiseaseSummaryDocument` for many curies in one ES round-trip
- `GET /api/disease/counts?ids=...` - gene/model/allele annotation counts per disease curie
- `GET /api/disease/{id}/ancestors` - root-first ancestor chain (capped at depth 30, cycle-safe)

Both batch endpoints reject requests with more than 500 ids (`BadRequestException`) so a large `ids` list cannot blow up an ES query.

**Per @oblodgett's review on the previous PR (#1610):** the original `DiseaseTermStub` POJO has been replaced with the team's `@JsonView` pattern.

- New `PublicView.DiseaseAncestor` view class (standalone, not extending other views, so only the explicitly annotated fields serialize)
- New `OntologyTermAncestorMixIn` adds `@JsonView(PublicView.DiseaseAncestor.class)` to `getCurie()` and `getName()` — a Jackson MixIn because `OntologyTerm` lives in `curation_api` and we can't annotate the source directly
- Registered in `RestDefaultObjectMapper` via `addMixIn(OntologyTerm.class, ...)`
- `getDiseaseAncestors` returns `List<DOTerm>` annotated with `@JsonView(PublicView.DiseaseAncestor.class)` — JSON output is identical to before (`[{curie, name}, ...]`)
- `DiseaseTermStub.java` deleted

Supersedes the closed #1610 (which was based on the older pre-SCRUM-6057 stage). Rebased onto current `stage` — imports updated for moved classes (`SearchDAO` -> `api.es.dao`, etc).

## Test plan
- [ ] `curl 'http://stage.alliancegenome.org/api/disease/batch?ids=DOID:14330,DOID:1287'` returns a map of curie -> document
- [ ] `curl 'http://stage.alliancegenome.org/api/disease/counts?ids=DOID:14330,DOID:1287'` returns `{ DOID:14330: { genes, models, alleles }, ... }`
- [ ] `curl 'http://stage.alliancegenome.org/api/disease/DOID:14330/ancestors'` returns the ancestor chain root-first, with each entry containing **only** `curie` and `name` (no other fields)
- [ ] Sending 501 comma-separated ids to either batch endpoint returns HTTP 400
- [ ] Swagger UI at `/q/swagger-ui/` lists the new endpoints


agr_api/src/main/java/org/alliancegenome/api/service/DiseaseESService.java
```diff
@@ -478,6 +490,105 @@ public String getAT(String category, String diseaseID) {
...
+			if (DISEASE_ROOT.equals(current)) break;
+			if (doc.getParents() == null || doc.getParents().isEmpty()) break;
+			current = doc.getParents().iterator().next().getCurie();
+		}
+		java.util.Collections.reverse(chain);
+		return chain;
+	}
+
+	private static final SearchDAO COUNTS_SEARCH_DAO = new SearchDAO();
+	private static final SearchDAO BATCH_TERMS_SEARCH_DAO = new SearchDAO();
```
oblodgett: shouldn't need two SearchDAO's here plus we typically put them at the top of the file.

agr_api/src/main/java/org/alliancegenome/api/service/DiseaseESService.java
```diff
@@ -478,6 +490,105 @@ public String getAT(String category, String diseaseID) {
 		return result;
 	}
 
+	private static final String DISEASE_ROOT = "DOID:4";
```
oblodgett: These constants should go to the top of the file.

jdepons:
Addressed all review feedback in subsequent commits — the earlier review comments describe the original code, not the current branch tip.

**Commit `1e21952b9` (`KANBAN-1322 address review feedback`):**
- @oblodgett `DiseaseESService.java:493` ("constants should go to the top of the file") — `DISEASE_ROOT`, `MAX_ANCESTOR_DEPTH`, and `COUNT_CATEGORIES` are now declared at the top of the class.
- @oblodgett `DiseaseESService.java:514` ("shouldn't need two SearchDAO's here") — `COUNTS_SEARCH_DAO` and `BATCH_TERMS_SEARCH_DAO` consolidated into a single `SEARCH_DAO` at the top of the class.
- Non-deterministic parent in `getAncestors()` — now picks the lexicographically-smallest parent curie at each step, so the breadcrumb is stable across requests in the Disease Ontology's DAG.
- N+1 ES queries in `getAncestors()` — left as-is (one round-trip per ancestor level, capped by `MAX_ANCESTOR_DEPTH = 30`), with a comment documenting the trade-off. Open to a follow-up if traffic warrants pre-materializing the ancestor path at index time.

**Commit `b6c62b511` (`KANBAN-1322 checkstyle fixes`):**
- Reviewdog `NeedBracesCheck` violations in `DiseaseController.java` and `DiseaseESService.java` — added explicit braces to all bare `if`/`for`/`while` bodies.

oblodgett (approved):
Do you know

PR #1611, opened 2026-05-19
KANBAN-678 add reference-scoped data endpoints for reference report

jdepons:
## Summary
New endpoints powering the Reference Report UI (matching `feature/KANBAN-678` in agr_ui):

Subject tables:
- `GET /api/reference/{id}/genes`
- `GET /api/reference/{id}/alleles`
- `GET /api/reference/{id}/models`

Annotation tables (paginated):
- `GET /api/reference/{id}/disease-annotations`
- `GET /api/reference/{id}/phenotype-annotations`
- `GET /api/reference/{id}/expression-annotations`
- `GET /api/reference/{id}/molecular-interactions`
- `GET /api/reference/{id}/genetic-interactions`
- `GET /api/reference/{id}/orthology`

Related papers:
- `GET /api/reference/{id}/related-papers?limit=10&includeOrthologs=false` - Jaccard-ranked related papers based on shared gene sets, with optional one-hop ortholog expansion

The expression / molecular / genetic interaction endpoints accept an optional `crossReferences` query param so callers can include alternate publication curies (PMID/MOD) in the lookup set.

## Test plan
- [ ] `curl 'http://stage.alliancegenome.org/api/reference/AGRKB:101000000828456/genes'` returns distinct genes for that reference
- [ ] Same with `/alleles`, `/models`, `/disease-annotations`, `/phenotype-annotations`, `/expression-annotations`, `/molecular-interactions`, `/genetic-interactions`, `/orthology`
- [ ] `curl 'http://stage.alliancegenome.org/api/reference/{id}/related-papers'` returns a ranked list with `sharedGenes`, `jaccard`, `sharedSpecies`
- [ ] `?crossReferences=PMID:12345,WB:WBPaper000xxx` on the cross-reference-aware endpoints expands the lookup set
- [ ] Reference detail page in the matching UI PR renders every section's table

Generated with Claude Code (https://claude.com/claude-code)

PR #1613, opened 2026-05-19
KANBAN-678 add reference-scoped data endpoints for reference report

jdepons:
## Summary
New endpoints powering the Reference Report UI (matching `feature/KANBAN-678` in agr_ui):

Subject tables:
- `GET /api/reference/{id}/genes`
- `GET /api/reference/{id}/alleles`
- `GET /api/reference/{id}/models`

Annotation tables (paginated):
- `GET /api/reference/{id}/disease-annotations`
- `GET /api/reference/{id}/phenotype-annotations`
- `GET /api/reference/{id}/expression-annotations`
- `GET /api/reference/{id}/molecular-interactions`
- `GET /api/reference/{id}/genetic-interactions`
- `GET /api/reference/{id}/orthology`

Related papers:
- `GET /api/reference/{id}/related-papers?limit=10&includeOrthologs=false` - Jaccard-ranked related papers based on shared gene sets, with optional one-hop ortholog expansion

The expression / molecular / genetic interaction endpoints accept an optional `crossReferences` query param so callers can include alternate publication curies (PMID/MOD) in the lookup set.

Supersedes the closed #1611 (which was based on the older pre-SCRUM-6057 stage). Rebased onto current stage — `SearchDAO`, `Pagination`, `JsonResultResponse`, `RestError*` import paths updated; annotation Document classes import from `core.document` (where SCRUM-6057 moved them) rather than the deleted `api.entity` package. Disease annotation deserialization still does the `APIServiceHelper.buildProvidersWithUrl` enrichment so the UI's Disease section Source column populates correctly.

## Test plan
- [ ] `curl 'http://stage.alliancegenome.org/api/reference/AGRKB:101000000828456/genes'` returns distinct genes for that reference
- [ ] Same with `/alleles`, `/models`, `/disease-annotations`, `/phenotype-annotations`, `/expression-annotations`, `/molecular-interactions`, `/genetic-interactions`, `/orthology`
- [ ] `/disease-annotations` response includes a populated `providers` field on each result
- [ ] `curl 'http://stage.alliancegenome.org/api/reference/{id}/related-papers'` returns a ranked list with `sharedGenes`, `jaccard`, `sharedSpecies`
- [ ] `?crossReferences=PMID:12345,WB:WBPaper000xxx` on the cross-reference-aware endpoints expands the lookup set
- [ ] Reference detail page in the matching UI PR renders every section's table

### DIFF
--- a/agr_api/src/main/java/org/alliancegenome/api/controller/LiteratureController.java
+++ b/agr_api/src/main/java/org/alliancegenome/api/controller/LiteratureController.java
@@ -96,11 +96,13 @@ public class LiteratureController implements LiteratureRESTInterface {
 																									String species,
 																									String asc) {
 		long startTime = System.currentTimeMillis();
+		List<String> xrefs = splitCommas(crossReferences);
+		requireCrossReferences(xrefs);
 		Pagination pagination = new Pagination(page, limit, sortBy, asc);
 		pagination.addFilterOption("geneExpressionAnnotation.expressionAnnotationSubject.taxon.species.fullName.keyword", species);
 		validate(pagination);
 		try {
-			return timed(referenceDataESService.getExpressionAnnotations(splitCommas(crossReferences), pagination), startTime);
+			return timed(referenceDataESService.getExpressionAnnotations(xrefs, pagination), startTime);
 		} catch (Exception e) {
 			throw restError("Error while retrieving expression annotations by reference", e);
 		}
@@ -115,11 +117,13 @@ public class LiteratureController implements LiteratureRESTInterface {
 																									String species,
 																									String asc) {
 		long startTime = System.currentTimeMillis();
+		List<String> xrefs = splitCommas(crossReferences);
+		requireCrossReferences(xrefs);
 		Pagination pagination = new Pagination(page, limit, sortBy, asc);
 		pagination.addFilterOption("geneMolecularInteraction.geneAssociationSubject.taxon.species.fullName.keyword", species);
 		validate(pagination);
 		try {
-			return timed(referenceDataESService.getMolecularInteractions(splitCommas(crossReferences), pagination), startTime);
+			return timed(referenceDataESService.getMolecularInteractions(xrefs, pagination), startTime);
 		} catch (Exception e) {
 			throw restError("Error while retrieving molecular interactions by reference", e);
 		}
@@ -134,11 +138,13 @@ public class LiteratureController implements LiteratureRESTInterface {
 																								String species,
 																								String asc) {
 		long startTime = System.currentTimeMillis();
+		List<String> xrefs = splitCommas(crossReferences);
+		requireCrossReferences(xrefs);
 		Pagination pagination = new Pagination(page, limit, sortBy, asc);
 		pagination.addFilterOption("geneGeneticInteraction.geneAssociationSubject.taxon.species.fullName.keyword", species);
 		validate(pagination);
 		try {
-			return timed(referenceDataESService.getGeneticInteractions(splitCommas(crossReferences), pagination), startTime);
+			return timed(referenceDataESService.getGeneticInteractions(xrefs, pagination), startTime);
 		} catch (Exception e) {
 			throw restError("Error while retrieving genetic interactions by reference", e);
 		}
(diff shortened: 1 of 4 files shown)


jdepons:
Addressed the review feedback in subsequent commits — the earlier review describes the original code, not the current branch tip.

**Commit `af8b08d07` (`KANBAN-678 require crossReferences on expression/interaction endpoints`):**
- The `id` path param was indeed dead on the expression / molecular-interaction / genetic-interaction endpoints. Stage ES indexes those document types by PMID/MOD curie, not by AGRKB curie, so `id` couldn't be used as a fallback filter. The controller now rejects requests missing a non-empty `crossReferences` with HTTP 400, and the OpenAPI `@Parameter` annotations are marked `required = true`. No more silent unscoped queries.

**Commit `86b62b9ca` (`KANBAN-678 checkstyle fixes`):**
- Reformatted multi-line method signatures in `LiteratureController` to pure tab indentation (the old tab+space visual alignment failed `RegexpSinglelineJavaCheck`).
- Added explicit braces to all bare `if` bodies (`NeedBracesCheck`).
- Empty record body `Scored {}` → `Scored { }` (`WhitespaceAroundCheck`).

agr_api/src/main/java/org/alliancegenome/api/service/ReferenceDataESService.java
oblodgett, line 1: I think it would be better to put all this functionality under the Literature ES Service. Reason: The "Literature" chain of documents can be followed back through to the ABC. We kind of make that distinction to know if we are talking about persistent store data (Referencess) or ABC data (Literature). Also some of the Generatic methods could be moved to the superclass ESService.

agr_api/src/main/java/org/alliancegenome/api/service/ReferenceDataESService.java
```diff
@@ -0,0 +1,447 @@
...
+			query.must(termsQuery("geneGeneticInteraction.evidence.referenceID.keyword", crossReferenceCuries));
+		}
+		return runQuery(query, pagination, hit -> mapHit(hit, GeneGeneticInteractionDocument.class));
+	}
+
+	public JsonResultResponse<Map<String, Object>> getGenesByReference(String referenceCurie) {
+		return getDistinctSubjects(GENE_SUBJECT_CATEGORIES, referenceCurie);
+	}
+
+	private static final List<String> GENE_AND_ALLELE_ANNOTATION_CATEGORIES = List.of(
```
oblodgett: Move to top of file.

agr_api/src/main/java/org/alliancegenome/api/service/ReferenceDataESService.java
```diff
@@ -0,0 +1,447 @@
...
+
+		ret.setTotal(subjects.size());
+		ret.setResults(subjects);
+		return ret;
+	}
+
+	@FunctionalInterface
+	private interface HitMapper<T> {
+		T map(SearchHit hit);
+	}
```
oblodgett: Im not sure about adding a FunctionalInterface, we use a global object mapper that can be injected. This kinda breaks from how we handle it everywhere else.

agr_api/src/main/java/org/alliancegenome/api/controller/LiteratureController.java
```diff
@@ -22,4 +40,231 @@ public LiteratureSummaryDocument getLiterature(String id) {
 			return literatureSummary;
 		}
 	}
+
+	@Override
+	public JsonResultResponse<DiseaseAnnotationDocument> getDiseaseAnnotationsByReference(
```
cmpich, line 45: I am puzzled what this method is returning: It's basically returning all DiseaseAnnotation filtered by general fields but not by any reference ID. SO, I am not sure what this endpoint is used for in the literature code.

agr_api/src/main/java/org/alliancegenome/api/service/ReferenceDataESService.java
jdepons: Moving stuff to the superclass is not an issue. Not sure I agree with the chain of documents thing.  The data I am showing is all persistent store data just keyed by the reference curie (ie gene disease annotations,  orthology, alleles phenotype annotations, etc).  It's really a view of what is in our persistent store for any individual reference.  Are you suggesting all that stuff be indexed in the Literature ES?  I'll bring it up tomorrow at backlog refinement and can add it to developer coordination on Monday if needed.

agr_api/src/main/java/org/alliancegenome/api/controller/LiteratureController.java
jdepons: The id does get applied — it's passed through to ReferenceDataESService.getDiseaseAnnotations, which filters on references.curie.keyword = id as a required clause. The pagination filters in the controller are just the table's optional dropdowns (species, disease name, etc.) layered on top. Endpoint backs the Disease Associations table on /reference/{id}.

agr_api/src/main/java/org/alliancegenome/api/service/ReferenceDataESService.java
jdepons: Dropped the HitMapper interface and the lambda passing runQuery. Kept mapHit on ESService per your superclass suggestion. The three uniform endpoints go through a small runTypedQuery helper now.  Disease and phenotype inline their loops since they need by category dispatch.

oblodgett (approved):

PR #1629, opened 2026-06-24
KANBAN-1385: count distinct positive genes for batch /disease/counts

Joe-Wdz:
KANBAN-1385: Refine gene counts at the top of portal pages


jdepons (approved):

agr_api/src/main/java/org/alliancegenome/api/service/DiseaseESService.java
```diff
@@ -631,15 +631,26 @@ public long countDistinctPositiveGenes(String diseaseID) {
 	}
 
 	private java.util.Map<String, Long> countByDisease(String category, String[] diseaseIds) {
```
oblodgett, line 633: @Joe-Wdz i would call this boolean "genesQuery" as that reflects the intent.

PR #1630, opened 2026-06-29
Removed zfin HTP variants

oblodgett:


cmpich (approved):
I am not sure why line 50 is marked as changed. It should not and it does not look like. Maybe a newline?

PR #1635, opened 2026-07-10
Add agr_api application.properties for local Quarkus config

kthorat-prog:
The uncommitted change in DiseaseRibbonService.java (getParentIDsFromStream) fixes a bug where a disease term's own ID wasn't being included in its ancestor set — it
  filtered a stream to only IDs contained in getDOParentTermIDs(doID), dropping doID itself unless it happened to be its own parent. The fix explicitly keeps          
  id.equals(doID) and adds a null-check on ancestorIDs to avoid an NPE if there are no parent terms.


oblodgett (changes requested):
We don't commit this file. If you want to make changes to the one that is used in production you make changes to the application.properties.defaults

kthorat-prog:
Re: the description mismatch noted above — that referenced a `DiseaseRibbonService.getParentIDsFromStream` fix (ensuring a term's own ID counts as its own ancestor) that predated this PR. Checked whether it's still needed: the old Neo4j-based `DiseaseRibbonService`/`DiseaseRepository` was removed entirely in the SCRUM-6057 ES migration, and the self-inclusion case is already handled there — both `DiseaseSummaryDocumentBuilder.buildSummaryDocument` (agr_curation) and `DoTermService`'s `closureMap` (agr_indexer) explicitly add the term's own curie to the parent closure set before it's persisted, with `DoTermService` even calling out the exact Neo4j-self-edge vs. `ontologytermclosure` gap in a comment. So no additional change needed — this is already handled upstream.

oblodgett (approved):

PR #1637, opened 2026-07-15
SCRUM-6204: embed CrossReference URL templates in publication docs

adamgibs:
Retire the client-side static resource-descriptor lookup. URL templates now flow curation -> indexer -> ES doc, so the UI builds linkouts directly and new resource descriptors take effect without a UI redeploy.

- Phenotype/Disease document publication fields shift from Set<String> to Set<CrossReference> (pubmedPublications, pubModPublications).
- Phenotype and Disease indexers walk Reference.crossReferences in ReferenceConstants.primaryXrefOrder for xref selection and attach the RDP on ExternalDatabaseReference-derived xrefs (OMIM/Orphanet).
- LiteratureIndexer enriches each Reference xref via the new ResourceDescriptorService (loaded once from curation).
- Controller filter paths and ExpressionFileGenerator updated to the nested field shape.


agr_indexer/src/main/java/org/alliancegenome/indexer/interfaces/ResourceDescriptorPageInterface.java
```diff
@@ -0,0 +1,30 @@
...
+import jakarta.ws.rs.POST;
+import jakarta.ws.rs.Path;
+import jakarta.ws.rs.Produces;
+import jakarta.ws.rs.QueryParam;
+import jakarta.ws.rs.core.MediaType;
+
+@Path("/resourcedescriptorpage")
+@Produces(MediaType.APPLICATION_JSON)
+@Consumes(MediaType.APPLICATION_JSON)
+public interface ResourceDescriptorPageInterface {
```
oblodgett: @adamgibs you should be able to use the curation one? vs creating this?

adamgibs: Yeah, I'll need to add a new endpoint to curation first

oblodgett (approved):

PR #1641, opened 2026-07-17
SCRUM-6253: add public /api/resourceDescriptors endpoint

adamgibs:
Index Resource Descriptors from the curation API into the site index and expose them via a Swagger-discoverable endpoint, replacing the resourceDescriptors.yaml in agr_schemas. Results are sorted by prefix.


agr_java_core/src/main/java/org/alliancegenome/core/document/ResourceDescriptorDocument.java
```diff
@@ -0,0 +1,25 @@
...
+
+import lombok.Data;
+
+@Data
+@JsonView({PublicView.ResourceDescriptor.class})
+public class ResourceDescriptorDocument extends ESDocument {
+	{
+		category = "resource_descriptor";
+	}
+	private String prefix;
```
oblodgett: instead of remapping all the fields which means maintenance if it changes, just put the curation object directly inside the document this saves lots of code and when it changes you don't have to touch the indexer code

oblodgett (approved):

PR #1649, opened 2026-07-28
KANBAN-864: add filter.referenceCitation to report-table endpoints

jimflybase:
## Summary

Adds a `filter.referenceCitation` query param to the report-table endpoints backing the new local AGRKB **Reference** column in the UI. It registers `references.shortCitation` as a filter option **without** a `.keyword` suffix, so `ESService` applies a contains (wildcard) match — as opposed to the existing `filter.reference`, which is an exact `.keyword` match on `pubmedPublications.referencedCurie`.

**Paired UI PR: alliance-genome/agr_ui#1836 — these must land together.** The UI sends `filter.referenceCitation`; without this PR, JAX-RS silently drops the unknown query param and the UI's Reference-column filter is ignored (the table returns every row and merely de-emphasizes non-matching citations).

## Endpoints covered

Each also covers its `/download` variant:

| endpoint | commit |
| --- | --- |
| `GET /api/disease/{id}/genes` | `968abffa4` |
| `GET /api/disease/{id}/alleles` | `968abffa4` |
| `GET /api/disease/{id}/models` | `968abffa4` |
| `GET /api/allele/{id}/diseases` | `968abffa4` |
| `GET /api/gene/{id}/phenotypes` | `968abffa4` |
| `GET /api/allele/{id}/phenotypes` | `f1760a211` |
| `POST /api/disease` (gene-page disease ribbon) | `ca17411fe` |

### Note on `/api/allele/{id}/phenotypes` (`f1760a211`)

This one was missed in the first pass and is worth a reviewer's attention. The UI reaches the allele-page Phenotypes table through the *same* shared `phenotypeTable.jsx` component as the gene page, but a *different* controller method (`AlleleController.getPhenotypePerAllele`). Because unknown query params are dropped silently, the filter appeared to work while actually returning unfiltered rows. Lesson for future filter work here: audit by endpoint, not by UI component.

### Note on `POST /api/disease` (`ca17411fe`)

The param is threaded through the public ribbon method into the private overload that carries `includePrimaryAnnotations`, so both the table and its download honor it.

## Verification

Run against a local `agr_api` (`quarkus:dev`) on stage ES.

| request | before | after |
| --- | --- | --- |
| `allele/RGD:7241046/phenotypes?filter.referenceCitation=2014` | 14 (unfiltered) | 11 |
| `allele/RGD:7241046/phenotypes?filter.referenceCitation=<nonsense>` | 14 (unfiltered) | 0 |
| `allele/RGD:7241046/phenotypes/download?filter.referenceCitation=2014` | 14 rows | 11 rows |
| `POST disease?focusTaxonId=NCBITaxon:9606` `["HGNC:11998"]` + `=2014` | 85 (unfiltered) | 8 |
| same + `=Bally` | 85 (unfiltered) | 2 |
| same + `=<nonsense>` | 85 (unfiltered) | 0 |
| `POST disease/download` + `=Bally` | 101 rows | 2 rows |

The five original endpoints were verified in an earlier session: genes 3312→14, alleles 432→23, models 1517→83, allele/diseases 5→3, gene/phenotypes 34→11; downloads honor the filter.

## Known issue (pre-existing, not introduced here)

`filter.reference` regressed from substring to exact matching in this release (SCRUM-6204). Not addressed here: the UI is keeping the PMID "Reference ID" column and its `filter.reference` behavior for now, since some users may want PMID filterability.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/agr_api/src/main/java/org/alliancegenome/api/controller/AlleleController.java
+++ b/agr_api/src/main/java/org/alliancegenome/api/controller/AlleleController.java
@@ -172,7 +172,7 @@ public class AlleleController implements AlleleRESTInterface {
 		Integer limit,
 		Integer page,
 		String sortBy,
-		String asc) {
+		String asc, String referenceCitation) {
 
 		LocalDateTime startDate = LocalDateTime.now();
 		Pagination pagination = new Pagination(page, limit, sortBy, asc);
@@ -182,6 +182,7 @@ public class AlleleController implements AlleleRESTInterface {
 		pagination.addFilterOption("generatedRelationString.keyword", associationType);
 		pagination.addFilterOption("diseaseQualifiers.keyword", diseaseQualifier);
 		pagination.addFilterOption("pubmedPublications.referencedCurie", filterReference);
+		pagination.addFilterOption("references.shortCitation", referenceCitation);
 		pagination.addFilterOption("primaryAnnotations.dataProvider.sourceOrganization.abbreviation", filterSource);
 
 
@@ -219,7 +220,7 @@ public class AlleleController implements AlleleRESTInterface {
 												Integer limit,
 												Integer page,
 												String sortBy,
-												String asc) {
+												String asc, String referenceCitation) {
 		JsonResultResponse<AlleleDiseaseAnnotationDocument> response = getDiseasePerAllele(alleleID,
 			filterOptions,
 			filterReference,
@@ -234,7 +235,7 @@ public class AlleleController implements AlleleRESTInterface {
 			150000,
 			page,
 			sortBy,
-			asc);
+			asc, referenceCitation);
 		Response.ResponseBuilder responseBuilder = Response.ok(diseaseToTdfTranslator.getAllRowsForAlleleDiseaseAnnotations(response.getResults()));
 		String alleleSymbol = getAllele(alleleID).getAllele().getAlleleSymbol().getFormatText();
 		APIServiceHelper.setDownloadHeaderByName(alleleID, alleleSymbol, EntityType.ALLELE, EntityType.DISEASE, responseBuilder);
--- a/agr_api/src/main/java/org/alliancegenome/api/controller/AlleleController.java
+++ b/agr_api/src/main/java/org/alliancegenome/api/controller/AlleleController.java
@@ -120,11 +120,13 @@ public class AlleleController implements AlleleRESTInterface {
 		String phenotype,
 		String source,
 		String reference,
-		String sortBy) {
+		String sortBy,
+		String referenceCitation) {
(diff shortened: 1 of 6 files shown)


oblodgett (approved):

oblodgett:
@jimflybase you need to make sure this branch is directly off an updated stage. It looks like your stage is old. You can just rebase this onto the end of stage and the conflicts should go away.

jimflybase:
Rebased onto the current `stage` (`cee67e867`) and force-pushed — the branch is now 3 commits ahead of `stage` and 0 behind. Thanks for catching it; the old base was 9 commits stale.

Two of the conflicts were worth more than a mechanical resolution, so noting how they were resolved:

**1. `pubmedPublications.referencedCurie` — took `stage`'s version, dropping the `.keyword` suffix my branch was carrying.** The old base had `...referencedCurie.keyword`, which makes `filter.reference` an exact match; `stage` has since dropped the suffix to restore substring matching. Resolving in favour of my side would have silently reverted that fix, so `stage` wins on all 6 of those lines and my new `references.shortCitation` line is simply added alongside. This matters for us specifically because we've decided to keep the PMID "Reference ID" column and its `filter.reference` behaviour rather than retire it.

**2. `getCountsOfDiseaseAnnotationsByAllele` / `...ByModel` — took `stage`'s version wholesale.** `stage` replaced the old `getDiseaseAnnotations*` delegation with `diseaseESService.countDistinctPositiveAlleles/Models(diseaseID)`. My branch had only touched those call sites incidentally (an extra positional argument for the new param), so that edit is obsolete and was dropped.

The resulting diff vs `stage` is purely additive: the `referenceCitation` parameter plus one `addFilterOption("references.shortCitation", …)` per endpoint. No occurrence of `referencedCurie.keyword` remains in the diff, and `countDistinctPositive*` is intact.

Re-verified against a local `agr_api` on the rebased tree:

| request | result |
| --- | --- |
| `allele/RGD:7241046/phenotypes` unfiltered | 14 |
| same + `filter.referenceCitation=2014` | 11 |
| `gene/RGD:2129/phenotypes` + nonsense term | 0 |
| `disease/DOID:14330/genes` + `=Tang` | 2 (matches a manual scan of the unfiltered rows) |
| `POST /api/disease` (ribbon) + `=Bally` | 2 |
| `disease/DOID:14330/genes` + `filter.reference=12133586` | 1 — substring behaviour from `stage` preserved |

Substring semantics confirmed still in place after the rebase: mid-word fragments match (`illeneuve`, `euve`, `016`, `ave` each return the same counts as their full-word forms).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1656, opened 2026-07-30
Scrum 6327

kthorat-prog:
The fix                                                                                                                                                                                                                           
                                                                                                                                                                                                                                    
  One small helper in the base class, called at every point an entity's notes get embedded:                                                                                                                                         
                                                                                                                                                                                                                                    
  // Indexer.java                                                                                                                                                                                                                   
  protected static void stripInternalNotes(List<Note> notes) {                                                                                                                                                                      
      if (CollectionUtils.isNotEmpty(notes)) {                                                                                                                                                                                      
          notes.removeIf(note -> Boolean.TRUE.equals(note.getInternal()));                                                                                                                                                          
      }                                                                                                                                                                                                                             
  }                                                                                                                                                                                                                                 
                                                                                                      Jump to bottom (click) ↓                                                                                                      
It mutates the list in place — removing internal notes from the actual object graph before that object gets handed to Jackson. Nothing else about the entity changes, so the entity/annotation itself still indexes and displays  
  normally; only the internal note content disappears.                                                                                                                                                                              
                                                                                                                                                                                                                                    
  Six call sites needed it, one per place a note-bearing entity gets embedded:                                                                                                                                                      
                                                                                                                                                                                                                                    
  1. GeneSummaryCurationIndexer — strips Gene.relatedNotes before indexing GeneSummaryDocument. This is the actual WBsf leak (WB:WBsf014559 etc. are Gene records with private_comment notes, confirmed by the DB pull).            
  2. DiseaseAnnotationCurationIndexer.populateBaseDiseaseAnnotationDocument — strips notes off the DiseaseAnnotation itself and its subject entity, right before dad.addPrimaryAnnotation(da). This is the fix for the 134 AGM      
  disease_note leaks (MGI:2656637 and the other 127 AGMs) — the annotation still gets added to the document, just without its note.                                                                                                 
  3. Same file, orthology rollup path and genetic-modifier / based-on-gene embedding — these are separate code paths that also embed a DiseaseAnnotation or a Gene/Allele/AGM directly, so they needed the same call to avoid       
  leaving a gap.                                                                                                                                                                                                                    
  4. PhenotypeAnnotationCurationIndexer — identical structure to disease annotations, same fix, defensive (currently zero internal notes there per the DB pull, but it's the same bug class).                                       
  5. TransgenicAlleleCurationIndexer and AlleleSummaryCurationIndexer — strip Allele.relatedNotes before embedding, also defensive (zero internal notes on alleles today).                                                          
                                                                                                                                                                                                                                    
  The 98 orphaned notes from the DB pull aren't attached to any entity, so no indexer ever touches them — nothing to fix there; that's dead data for the curation team to clean up separately.  

Current test coverage for the WBsf tickets:                                                                                                                          
  - WB:WBsf014559 (original description) — single note case + new multi-note case                                                                                      
  - WB:WBsf019098, 019099, 019217, 019218 (comment 92477) — single note case 

WB:WBsf982337 added to GeneNoteStrippingTest's curie list, sabotage-checked, fix restored, full suite green (5/5). All 6 WBsf gene IDs from the ticket (description +
  comment 92477 + Olin's DB pull) are now covered:                                                                                                                     
                                                                                                                                                                       
  - WB:WBsf014559 — dedicated test with two distinct internal notes                                                                                                    
  - WB:WBsf019098, 019099, 019217, 019218, 982337 — single private_comment + public note case                                                                          
                                                                                                                                                                       
  Plus MGI:7596100 for the AGM disease-annotation leak path. Working tree is unchanged otherwise — just the fix files and the new agr_indexer/src/test/ directory. Let 
  me know if there's anything else to add before this gets committed.


agr_indexer/src/main/java/org/alliancegenome/indexer/indexers/Indexer.java
```diff
@@ -277,6 +279,18 @@ protected HashMap<String, Integer> buildSpeciesOrder(String taxonCurie) {
...
 
+	/**
+	 * Removes notes marked internal=true from a relatedNotes list before it's embedded in a
+	 * public ES document. Curation entities carry internal notes (e.g. private_comment) alongside
+	 * public ones with no server-side view filtering applied during indexing, so this must run on
+	 * every entity's relatedNotes before it reaches indexDocument(s) — see SCRUM-6327.
+	 */
+	protected static void stripInternalNotes(List<Note> notes) {
+		if (CollectionUtils.isNotEmpty(notes)) {
+			notes.removeIf(note -> Boolean.TRUE.equals(note.getInternal()));
```
oblodgett: @kthorat-prog I think we need to also strip obsolete as well... you might need to ask chris about it.

kthorat-prog: Yes, Sure will ask Chris about it. Thank you.

kthorat-prog: Chris confirmed to strip off Obsolete. I did the changes and pushed them. Please check.

oblodgett (approved):

